PR #72, opened 2022-12-15
Coverage Comment

wistuba:
Bot automatically reports coverage, see below or [full example](https://github.com/py-cov-action/python-coverage-comment-action-v3-example/pull/1).


wistuba:
This could be an alternative but I think this would need to be enabled by awslabs: https://about.codecov.io/

610v4nn1:
I checked codecov and it requires an extensive amount of permissions. Coveralls (https://coveralls.io/) seems to provide a comparable service, at least for our purposes, and it is significantly less invasive. It also needs to be enabled by awslabs.

610v4nn1 (approved):

PR #85, opened 2023-01-02
Downgrading sphinx-opengraph version after failure

610v4nn1:
Rolling back sphinx-opengraph upgrade after failure in doc build on RTD


wistuba (changes requested):
I've checked locally with 0.7.3, same problem. This must be coming from somewhere else. I'll have a look into it.

wistuba:
This was caused with the latest release of Sphinx 6 days ago and causes our custom theme to break. The library owners already addressed this problem and we will get a solution soon: https://github.com/pydata/pydata-sphinx-theme/pull/1097

PR #100, opened 2023-01-16
Add finetuning model updater

610v4nn1:
Add a model updater doing fine-tuning.
This will simplify usage by providing an ad-hoc model updater for the fine-tuning baseline instead of requiring the usage of the SimpleModelUpdater in combination with Learner.


test/renate/updaters/experimental/test_fine_tuning.py
```diff
@@ -0,0 +1,33 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
wistuba, line 1: you might want to add the learner to conftest.py instead

PR #102, opened 2023-01-17
Avalanche Integration

### LINKED ISSUE #113: Avalanche Integration
Make some Avalanche updaters available in Renate.

wistuba:
Closes #113 

Missing:
- Buffer augmentation
- Use of custom metrics. Only supports loss and classification error.


src/renate/updaters/avalanche/model_updater.py
```diff
@@ -0,0 +1,328 @@
...
+            )
+        results = self._learner.eval(benchmark.test_stream)
+        SyneTunePlugin.report(
+            epoch=self._max_epochs, results=results, report_val=val_dataset_exists
+        )
+
+        return self._model
+
+
+class ExperienceReplayAvalancheModelUpdater(AvalancheModelUpdater):
```
610v4nn1, line 201: I know this follows the same structure of the other model updaters but since in this case we will always have a fairly standard model updater instantiating a specific learner I wonder if we should not consider having an AvalancheModelUpdater -- which we already have -- that instantiates a learner based on an input parameter (kwargs for the extra arguments).
We can discuss this offline.

610v4nn1:
Great progress, thank you Martin!

src/renate/updaters/model_updater.py
```diff
@@ -61,11 +61,11 @@ class RenateModelCheckpoint(ModelCheckpoint):
     Args:
         model: Model to be saved when creating a checkpoint.
         state_folder: Checkpoint folder location.
-        val_enabled: Whether validation was enabled in the Learner. Forwarded to `SyneTuneCallback`.
+        val_enabled: Whether validation was enabled in the Learner. Forwarded to `SyneTunePlugin`.
```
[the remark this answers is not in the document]
wistuba: auto refactoring. will fix.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 numpy>=1.17.2, <1.24.2
-torch>=1.10.0, <1.13.2
+torch>=1.10.0, <1.12.2
```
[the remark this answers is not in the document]
wistuba: checkpointing in avalanche won't work with 1.13: https://github.com/ContinualAI/avalanche/issues/1270

src/renate/benchmark/models/base.py
```diff
@@ -0,0 +1,63 @@
...
+import torch
+from torch import nn
+
+from renate import defaults
+from renate.models import RenateModule
+from renate.models.prediction_strategies import ICaRLClassificationStrategy, PredictionStrategy
+
+
+# TODO: merge unit tests for the submodules
+class RenateBenchmarkingModule(RenateModule, ABC):
```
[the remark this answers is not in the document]
wistuba: It was meant to share code between our models but since we decided to move the iCaRL logic to the RenateModule directly, it features something more right now. Will add more documentation.

src/renate/defaults.py
```diff
@@ -131,8 +138,12 @@ def model_file(state_folder: str):
     return os.path.join(state_folder, "model.ckpt")
 
 
+LEARNER_CHECKPOINT_NAME = "learner.ckpt"
+AVALANCHE_CHECKPOINT_NAME = "avalanche.ckpt"
```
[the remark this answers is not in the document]
wistuba: it was used directly. I've changed it to match our current usage

src/renate/models/renate_module.py
```diff
@@ -53,7 +53,11 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         loss_fn: The loss function to be optimized during the training.
     """
 
-    def __init__(self, constructor_arguments: dict, loss_fn: torch.nn.Module):
```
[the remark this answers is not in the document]
wistuba: used to have more args which black formatted with trailing comma. unfortunately, black is not able to undo it. fixed it manually

src/renate/memory/buffer.py
```diff
@@ -201,7 +200,12 @@ def set_transforms(
         self._target_transform = target_transform
 
     def to_tensors(self) -> Tuple[Tensor, ...]:
```
lballes, line 202: This might lead to errors for _general_ inputs. `Buffer._data_points` is a tuple `(inputs, outputs)`, where `outputs` is a `torch.Tensor`, but `inputs` may be `torch.Tensor` or a tuple/dict thereof. I guess that the avalanche learners only support the simple `(X, y)` case? If so, we should put a check here, making sure that `inputs` is `torch.Tensor`.

lballes (changes requested):

src/renate/memory/buffer.py
wistuba: will this be a problem as long as we won't change the `ModelUpdater` for training a model?

lballes: I don't understand your comment. My point is: If your `Buffer` contains general inputs, say `buffer[0] = (x, y)` with `x = {"token_ids": aaa, "attention_mask": bbb}` and you call `Buffer.to_tensors()`, this will lead to an error. It would be easy to make the method more general by calling it recursively.

src/renate/utils/avalanche.py
```diff
@@ -1,55 +1,50 @@
...
-    """Helper class to use dataset subsets with Avalanche.
+class AvalancheDataset(Dataset):
+    """A Dataset consumable by Avalanche updaters."""
 
-    Since Avalanche does not support the Subset class, this class is used to convert subsets.
-    Furthermore, some Avalanche updaters (e.g. iCaRL) directly access the targets attribute.
+    def __init__(self, inputs: Inputs, targets: List[int]):
+        self._inputs = inputs
+        self._targets = targets
+        self.targets = torch.tensor(targets, dtype=torch.long)
```
lballes, line 23: Why is this needed? Is it used by `avalanche`?

wistuba: some avalanche updaters directly access `targets` in the dataset

lballes (approved):

PR #121, opened 2023-02-07
Renaming arguments and functions to simplify usage

610v4nn1:
Renaming the function used to execute the training job and a few arguments to make its usage more intuitive


src/renate/run/training_job.py
```diff
@@ -227,7 +227,7 @@ def _get_transfer_learning_task_evaluations(
     metric: str,
     max_epochs: int,
 ) -> Optional[TransferLearningTaskEvaluations]:
-    """Converts data frame with tuning results of a single update step into
+    """Converts data frame with run results of a single update step into
```
wistuba, line 230: training results?

src/renate/run/training_job.py
```diff
@@ -337,7 +337,7 @@ def _load_tuning_history(
 def _merge_tuning_history(
     new_tuning_results: pd.DataFrame, old_tuning_results: Optional[pd.DataFrame] = None
 ) -> pd.DataFrame:
-    """Merges old tuning history with tuning results from current chunk.
+    """Merges old tuning history with run results from current chunk.
```
wistuba, line 340: training results?

src/renate/run/training_job.py
```diff
@@ -140,9 +140,9 @@ def execute_tuning_job(
         backend in defaults.SUPPORTED_BACKEND
     ), f"Backend {backend} is not in {defaults.SUPPORTED_BACKEND}."
     if backend == "local":
-        return _execute_tuning_job_locally(
-            state_url=state_url,
-            next_state_url=next_state_url,
+        return _execute_tuning_and_tuning_job_locally(
```
wistuba, line 143: a bit odd and lengthy name

examples/simple_classifier_cifar10/start_with_hpo.py
```diff
@@ -4,7 +4,7 @@
 from syne_tune.backend.sagemaker_backend.sagemaker_utils import get_execution_role
 from syne_tune.config_space import choice, loguniform, uniform
 
-from renate.tuning import execute_tuning_job
+from renate.run import run_training_job
```
wistuba, line 7: what about `training` rather than `run`? We could also call the function simply `train` or `update`.

610v4nn1: I thought about update but since we have the `updaters` package, it can be quite confusing.
`training` can be OK, but pretty much everything we do is about training.

I don't mind changing the name but I don't see great alternatives

src/renate/run/training_job.py
610v4nn1: I agree, it's a long name but I think it's good to be specific. 
Also, this is not exposed to the user.

examples/simple_classifier_cifar10/start_with_hpo.py
wistuba: We could also stick closer to the SM interface and create a `Renate` object on which we call `fit`. Since many arguments in SM and Renate are similar, might be a convenient interface for the experience SM user.

610v4nn1: Something like a `RenateEstimator`?

wistuba: In SM it is `sagemaker.pytorch.estimator.PyTorch` (no `Estimator`). That solution also has the advantage that we can potentially use the object to access specific properties of the job (model, hpo history etc)

src/renate/run/training_job.py
lballes: Why is it tuning _and_ tuning job? I don't understand this.

src/renate/benchmark/experimentation.py
```diff
@@ -301,16 +301,16 @@ def _execute_experiment_job_locally(
...
             config_space=config_space,
             metric=metric,
             backend="local",
             updater=config_space["updater"],
             max_epochs=config_space["max_epochs"],
             chunk_id=update_id,
-            state_url=state_url,
-            next_state_url=next_state_url,
+            input_state_url=state_url,
+            output_state_url=next_state_url,
```
lballes, line 313: We should be consistent with the names in all the functions. Can't we just call this `input_state_url` and `output_state_url` everywhere?

src/renate/cli/run_remote_job.py
```diff
@@ -1,15 +1,15 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 """
-This script is used to launch model tuning on SageMaker. Previously passed arguments on a different
+This script is used to launch model run on SageMaker. Previously passed arguments on a different
 machine are loaded again and the update process is started with these parameters. Arguments expected
-are described in renate.tuning.execute_tuning_job.
+are described in renate.run.execute_tuning_job.
```
lballes, line 6: This should be `run_training_job`, right?

src/renate/training/training.py
```diff
@@ -502,9 +502,9 @@ def _execute_tuning_job_locally(
     accelerator: str,
     devices: int,
 ):
-    """Executes the tuning job locally.
+    """Executes the training job locally.
 
-    See renate.tuning.execute_tuning_job for a description of arguments.
+    See renate.run.run_training_job for a description of arguments.
```
lballes, line 507: Is this still the correct path? I guess it should be `renate.training.`?

src/renate/training/training.py
```diff
@@ -610,9 +610,9 @@ def submit_remote_job(
     job_name: str,
     **job_kwargs: Any,
 ) -> str:
-    """Executes the tuning job on SageMaker.
+    """Executes the training job on SageMaker.
 
-    See renate.tuning.execute_tuning_job for a description of arguments."""
+    See renate.run.run_training_job for a description of arguments."""
```
lballes, line 615: `renate.training.x`?

lballes (changes requested):

lballes (approved):

PR #127, opened 2023-02-14
Permutation Bug Fix

wistuba:
Permutation scenarios had a bug that cause all permutations but first to be the same.

Relaxing sum to 1 check to account for problems with precision.


610v4nn1 (changes requested):
LGTM but can we add a test?

610v4nn1:
There are a couple of flake error, but the changes look good.

610v4nn1 (approved):

PR #128, opened 2023-02-16
Allow buffer to store (nested) tuples/dictionaries of tensors

lballes:
Allow buffer to store (nested) tuples/dictionaries of tensors

### DIFF
--- a/src/renate/memory/buffer.py
+++ b/src/renate/memory/buffer.py
@@ -11,10 +11,9 @@ from renate import defaults
 from renate.utils.pytorch import get_generator
 
 
-DataTuple = Tuple[torch.Tensor]
 DataDict = Dict[Hashable, torch.Tensor]
-NestedTensors = Union[torch.Tensor, DataTuple, DataDict]  # TODO: Technically, they could be nested.
-Index = int  # TODO: Include slices in type hint
+NestedTensors = Union[torch.Tensor, Tuple[torch.Tensor], Dict[Hashable, torch.Tensor]]
+Index = Union(int, slice)
 
 
 def _make_storage(data_point: NestedTensors, length: int) -> NestedTensors:
@@ -163,13 +162,13 @@ class DataBuffer(Dataset, ABC):
                 targets = self._target_transform(targets)
             return (inputs, targets), metadata
 
-    def __setitem__(self, idx: int, data_and_metadata: Tuple[NestedTensors, NestedTensors]) -> None:
+    def __setitem__(self, idx: int, data_and_metadata: Tuple[NestedTensors, DataDict]) -> None:
         """Replaces a data point in the buffer."""
         data, metadata = data_and_metadata
         _insert_data_point(self._data_points, idx, data)
         _insert_data_point(self.metadata, idx, metadata)
 
-    def _append(self, data: NestedTensors, metadata: NestedTensors) -> None:
+    def _append(self, data: NestedTensors, metadata: DataDict) -> None:
         """Appends a data point to the internal storage."""
         if not len(self):
             self._data_points = _make_storage(data, self._max_size)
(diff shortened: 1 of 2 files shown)


src/renate/memory/buffer.py
```diff
@@ -231,25 +258,28 @@ def __init__(
...
+            # Extend the size of the storage by a factor of 2.
+            current_data_points = self._data_points
+            self._data_points = _make_storage(
+                _get_data_point(current_data_points, 0), self._max_size * 2
+            )
+            _insert_data_point(self._data_points, slice(self._max_size), current_data_points)
+            current_metadata = self.metadata
+            self.metadata = _make_storage(_get_data_point(current_metadata, 0), self._max_size * 2)
+            _insert_data_point(self.metadata, slice(self._max_size), current_metadata)
             self._max_size *= 2
```
wistuba, line 283: nit: would update `_max_size` first and then use it to create the new storage.

test/renate/memory/test_buffer.py
```diff
@@ -12,19 +12,74 @@
...
     SlidingWindowBuffer,
 )
+from renate.memory.buffer import _make_storage, _insert_data_point, _get_data_point
+
+
+class NestedTensorDataset(torch.utils.data.Dataset):
+    def __init__(self, nested_tensors):
+        self._nested_tensors = nested_tensors
+        self._length = None
+        self._get_len(nested_tensors)
```
wistuba, line 22: I would assign this directly to `self._length` instead of setting it in `_get_len()`

test/renate/memory/test_buffer.py
```diff
@@ -144,16 +199,15 @@ def test_buffer_get_state_dict(max_size):
...
 
     for key in ["max_size", "storage_mode", "seed", "count", "data_points"]:
         assert getattr(buffer, "_" + key) == state_dict[key]
     assert buffer.metadata == state_dict["metadata"]
     assert state_dict["size"] == max_size
-    assert len(state_dict["data_points"]["0"]) == max_size
+    assert len(state_dict["data_points"][0]) == max_size
 
 
 @pytest.mark.parametrize("buffer_type", [ReservoirBuffer, SlidingWindowBuffer])
```
wistuba, line 213: maybe add independent tests for make_storage, add_data_point etc?

test/renate/memory/test_buffer.py
lballes: `_get_len()` is recursively checking all tensors for consistent lengths. I think this is the most concise way to do this.

test/renate/memory/test_buffer.py
lballes: There's an (admittedly rudimentary) `test_nested_tensor_storage` above, that tests these methods for different nested tensor structures.

test/renate/memory/test_buffer.py
wistuba: it's as concise but you get rid of the side effect and the getter actually returns a value:

```
    def _get_len(self, nested_tensors, expected_length=None) -> int:
        if isinstance(nested_tensors, torch.Tensor):
            length = nested_tensors.size(0)
            assert length == expected_length or expected_length is None
            return length
        elif isinstance(nested_tensors, tuple):
            for t in nested_tensors:
                expected_length = self._get_len(t, expected_length)
            return expected_length
        elif isinstance(nested_tensors, dict):
            for t in nested_tensors.values():
                expected_length = self._get_len(t, expected_length)
            return expected_length
```

lballes: Yep, fine with me. I changed it.

src/renate/memory/buffer.py
lballes: I don't really like that since I need the "old" `max_size` in the `_insert_data_point`.

wistuba (approved):

PR #129, opened 2023-02-16
Add data module for Huggingface text datasets

lballes:
Add data module for Huggingface text datasets


610v4nn1 (changes requested):

test/renate/benchmark/test_scenarios.py
```diff
@@ -7,7 +7,7 @@
 import torch
 from torchvision.transforms.functional import rotate
 
-from datasets import DummyTorchVisionDataModule, DummyTorchVisionDataModuleWithChunks
+from data import DummyTorchVisionDataModule, DummyTorchVisionDataModuleWithChunks
```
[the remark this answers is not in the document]
lballes: Renaming.

src/renate/benchmark/datasets/nlp_datasets.py
```diff
@@ -1,78 +1,120 @@
...
-        train_data = to_map_style_dataset(cls(root=self._data_path, split="train"))
-        self._train_data, self._val_data = self._split_train_val_data(train_data)
-        self._test_data = to_map_style_dataset(cls(root=self._data_path, split="test"))
+
+        def tokenize_fn(batch):
+            return self._tokenizer(batch[self._input_column], **self._tokenizer_kwargs)
+
+        columns = ["input_ids", "attention_mask", self._target_column]
+
+        self._train_data = self._train_data.map(tokenize_fn, batched=True)
```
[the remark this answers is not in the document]
lballes: Good point, I will make clear that this is for text datasets.

src/renate/benchmark/datasets/nlp_datasets.py
```diff
@@ -1,78 +1,120 @@
...
         data_path: the path to the folder containing the dataset files.
-        src_bucket: the name of the s3 bucket. If not provided, downloads the data from original
-            source.
-        src_object_name: the folder path in the s3 bucket.
-        dataset_name: Name of the torchvision dataset.
+        dataset_name: Name of the dataset, see https://huggingface.co/datasets.
+        input_column: Name of the column containing the input text.
+        target_column: Name of the column containing the target (e.g., class label).
+        tokenizer: Tokenizer to apply to the dataset.
+        tokenize_kwargs: Keyword arguments to be passed to the tokenizer.
```
610v4nn1, line 47: It would be good to provide a pointer to the list of the tokenizers and their kwargs.

src/renate/benchmark/datasets/nlp_datasets.py
```diff
@@ -1,78 +1,120 @@
...
+    will be used if present. Otherwise, a validation set may be split off of the training data
+    using the `val_size` argument.
 
     Args:
         data_path: the path to the folder containing the dataset files.
-        src_bucket: the name of the s3 bucket. If not provided, downloads the data from original
-            source.
-        src_object_name: the folder path in the s3 bucket.
-        dataset_name: Name of the torchvision dataset.
+        dataset_name: Name of the dataset, see https://huggingface.co/datasets.
```
610v4nn1, line 43: It would be good to specify that you support text datasets

test/renate/data/test_data_module.py
```diff
@@ -156,3 +132,48 @@ def test_tiny_imagenet_data_module(tmpdir):
...
+    ],
+)
+@pytest.mark.parametrize(
+    "tokenizer", [transformers.DistilBertTokenizer.from_pretrained("distilbert-base-uncased")]
+)
+@pytest.mark.parametrize(
+    "tokenizer_kwargs",
+    [
+        None,
+        dict(padding="max_length", max_length=128, truncation=True),
```
610v4nn1, line 150: When passing None, the default configuration will be used and the default is the same configuration you specify here. It would be good to run the test with two different configurations.

src/renate/benchmark/datasets/nlp_datasets.py
```diff
@@ -1,78 +1,120 @@
...
             seed=seed,
         )
-        if dataset_name not in TorchTextDataModule.dataset_dict:
-            raise ValueError(f"Dataset {self._dataset_name} currently not supported.")
         self._dataset_name = dataset_name
+        self._input_column = input_column
+        self._target_column = target_column
+        self._tokenizer = tokenizer
+        self._tokenizer_kwargs = tokenizer_kwargs or dict(
+            padding="max_length", max_length=128, truncation=True
```
610v4nn1, line 73: it would be good to move the default configuration of the tokenizer in the Renate defaults.

lballes: I don't think that's a good idea, since this is benchmarking code that has nothing to do with renate.

test/renate/data/test_data_module.py
lballes: Good point.

src/renate/benchmark/datasets/nlp_datasets.py
610v4nn1: I don't see that as a problem, but we can create a set of default for the benchmarks if you prefer

lballes: These defaults are specific to huggingface tokenizers and will only be used here. Is there really a good reason to outsource this to a separate file? If I'm a user of this class, I want to _see_ the defaults so that I can decide whether I need to change them.

lballes: Moved it to `defaults.py` after offline discussion

610v4nn1 (approved):

610v4nn1 (approved):

Issue #115, opened 2023-02-06
Simplify Benchmarking


wistuba:
Passing arguments when benchmarking can be improved. Arguments are too long, same values need to be passed multiple times and values are limited to primitive variable types.

PR #136, opened 2023-02-20
Convenient Use of Function Parameters

wistuba:
Addresses Part 1 of #115: Users can use arbitrary argument names for the functions and parameters are automatically forwarded based on the function interface.


src/renate/cli/parsing_functions.py
```diff
@@ -493,32 +638,96 @@ def get_transforms_kwargs(
...
+    "ER": _add_experience_replay_arguments,
+    "DER": _add_dark_experience_replay_arguments,
+    "POD-ER": _add_pod_experience_replay_arguments,
+    "CLS-ER": _add_cls_experience_replay_arguments,
+    "Super-ER": _add_super_experience_replay_arguments,
+    "GDumb": _add_gdumb_arguments,
+    "Joint": _add_joint_arguments,
+    "FineTuning": _add_finetuning_arguments,
+    "RD": _add_rd_learner_arguments,
+    "OfflineER": _add_replay_learner_arguments,
```
610v4nn1, line 732: Do we really need to maintain these functions? can we remove them in favor of an automatic detection of the arguments of each MU with get_function_args (or similar)?

610v4nn1:
It would be good to explain the usage of the arguments in the documentation

src/renate/cli/parsing_functions.py
wistuba: These functions extend the arguments parsed by argsparse. I don't see how we can remove this.

wistuba:
> It would be good to explain the usage of the arguments in the documentation

This is only the first part of a larger change that I break down into self-contained features to be better consumed by the reviewers. Explaining the current state won't make too much sense since it will change a bit. I'll add it with the next change.

src/renate/cli/parsing_functions.py
610v4nn1: As discussed offline, let's open a separate issue for this

PR #140, opened 2023-02-21
Exposing parameter to make trainer deterministic

610v4nn1:
The PTL Trainer output is not deterministic when the training is running on GPU. This can be problematic for users that want to reproduce results from scientific experiments or other kind of tests. It is possible to make the Trainer deterministic by using an ad-hoc flag but that impacts the training speed. 

With the changes in this PR we offer the possibility to make the Trainer deterministic. 


test/renate/updaters/test_model_updater.py
```diff
@@ -32,6 +33,39 @@ def test_simple_model_updater(tmpdir, provide_folder):
     assert not torch.allclose(y_hat_before_train, y_hat_after_train)
 
 
+def test_deterministic_updater():
+    # The behavior is always deterministic on CPU but it can become non-deterministic on GPU
```
wistuba, line 37: given that we use no GPU, this test is not checking what the name implies. I think we can keep a test like this but it would test whether seeding works (maybe we have this one already somewhere).

it would be sufficient to test whether `deterministic_trainer` is forwarded to `Trainer` since we can assume that Lightning is working. I'm not sure we have a good way to test that.

610v4nn1: I agree with your point, it's written also in the comment on lines 37 and 38.
I was more thinking that this can run on GPU with the integration tests.

src/renate/defaults.py
```diff
@@ -144,3 +145,5 @@ def metric_summary_file(logs_folder: str, special_str: str = ""):
 
 def hpo_file(state_folder: str):
     return os.path.join(state_folder, "hpo.csv")
+
+    return None
```
wistuba, line 149: remove

src/renate/updaters/model_updater.py
```diff
@@ -174,6 +178,7 @@ def __init__(
         logger: Logger = defaults.LOGGER(**defaults.LOGGER_KWARGS),
         accelerator: defaults.SUPPORTED_ACCELERATORS_TYPE = defaults.ACCELERATOR,
         devices: Optional[int] = None,
+        deterministic_trainer: bool = False,
```
wistuba, line 181: use the constant

test/renate/updaters/test_model_updater.py
wistuba: integration tests will not run the unit tests. for completeness, one would also need to check the False case

src/renate/updaters/model_updater.py
wistuba: why is this marked as resolved?

wistuba (approved):

PR #147, opened 2023-02-23
Add SoftSortingScenario

lballes:

### DIFF
--- a/src/renate/benchmark/scenarios.py
+++ b/src/renate/benchmark/scenarios.py
@@ -324,7 +324,7 @@ class SoftSortingScenario(Scenario):
         idx_ordered = torch.multinomial(
             sampling_probs, num_samples=len(dataset), replacement=False, generator=rng
         )
-        split = torch.split(idx_ordered, split_size_or_sections=self._num_tasks)
+        split = torch.tensor_split(idx_ordered, self._num_tasks)
         return [Subset(dataset, idx) for idx in split]
 
     def setup(self) -> None:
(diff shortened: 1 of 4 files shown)


src/renate/benchmark/scenarios.py
```diff
@@ -264,3 +264,76 @@ def setup(self) -> None:
...
+        seed: int = defaults.SEED,
+    ) -> None:
+        super().__init__(data_module, num_tasks, chunk_id, seed)
+        self._feature_idx = feature_idx
+        self._exponent = exponent
+
+    def _get_features(self, dataset: Dataset) -> torch.Tensor:
+        features = torch.empty(len(dataset))
+        for i in range(len(dataset)):
+            features[i] = dataset[i][self._feature_idx].mean()
```
610v4nn1, line 312: I think it would be good to write what's the expected input shape and maybe give an example with images

test/renate/benchmark/test_scenarios.py
```diff
@@ -197,3 +198,27 @@ def test_iid_scenario():
...
+            seed=data_module._seed,
+        )
+        scenario.prepare_data()
+        scenario.setup()
+        for stage in ["train", "val"]:
+            scenario_data = getattr(scenario, f"{stage}_data")()
+            for j in range(len(scenario_data)):
+                x, y = scenario_data[j]
+                assert x not in counter
+                counter[x] = 1
```
610v4nn1, line 223: This only checks for the absence of duplicates.
Can we check the sorting as well? maybe using an extremely large exponent?

src/renate/benchmark/scenarios.py
lballes: Did you see the doc string for `feature_idx`? I thought that was clear enough, but if you think there's something missing, I'll add that.

test/renate/benchmark/test_scenarios.py
lballes: I added a check for a decreasing _average_ value of the relevant feature between chunks.

src/renate/benchmark/scenarios.py
lballes: I updated it.

610v4nn1 (approved):

PR #149, opened 2023-02-24
Add memory-mapped tensor storage for nested tensor structures

lballes:
This is in preparation of a more scalable version of the memory buffers.


src/renate/memory/storage.py
```diff
@@ -0,0 +1,151 @@
...
+}
+
+
+def mmap_tensor(filename: str, shape: Tuple[int, ...], dtype: torch.dtype) -> torch.Tensor:
+    """Creates or accesses a memory-mapped tensor."""
+    t = torch.from_file(filename, shared=True, size=math.prod(shape), dtype=dtype, device="cpu")
+    return t.view(shape)
+
+
+def _create_or_access_storage(path: str, length: int, shapes, dtypes) -> NestedTensors:
```
lballes, line 26: There would be an alternative way to implement this, where instead of giving `shapes` and `dtypes`, we just give a single prototype data point. That would also make superfluous the `get_nested_tensor_schema` method and the need to convert dtypes to strings, et cetera. If you'd like that solution better, I could replace it.

wistuba (changes requested):
I would like to see the function outside the Storage class that deal with the NestedTensors to be part of the class in such a way that any class extending the Storage class will not need to deal with the NestedTensors but only need to write the function for Tensor.

Don't we want to have a class MemoryMappedStorage?

lballes:
> I would like to see the function outside the Storage class that deal with the NestedTensors to be part of the class in such a way that any class extending the Storage class will not need to deal with the NestedTensors but only need to write the function for Tensor.

I can make the functions part of `Storage`. I was hesitant to do that because they are recursive, but I guess there's nothing keeping us from doing that ;)
 
> Don't we want to have a class MemoryMappedStorage?

This is using memory-mapped tensors. I will rename to `MemoryMappedStorage`.

src/renate/memory/storage.py
lballes: @wistuba Thoughts on this?

lballes: I decided to go with this, see latest revision.

wistuba (approved):

PR #156, opened 2023-02-27
HueShiftScenario and FeatureSortingScenario Fix

wistuba:
Adding `HueShiftScenario` and fixes and refactors `FeatureSortingScenario`.

### DIFF
--- a/src/renate/benchmark/scenarios.py
+++ b/src/renate/benchmark/scenarios.py
@@ -6,7 +6,7 @@ from typing import Callable, List, Tuple, Union
 import numpy as np
 import torch
 from torch.utils.data import Dataset, Subset
-from torchvision.transforms import Lambda, RandomRotation
+from torchvision.transforms import Lambda, RandomRotation, ToPILImage
 
 from renate import defaults
 from renate.data.data_module import RenateDataModule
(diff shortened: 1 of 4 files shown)


src/renate/benchmark/scenarios.py
```diff
@@ -244,39 +244,79 @@ def __init__(
...
-    This extracts a feature value for each data point (see `feature_idx`) and computes sampling
-    probabilities proportional to `feature_value ** exponent`. An ordering of the data points is
-    then sampled from a corresponding multinomial distribution without replacement. This results in
-    a "soft" sorting, where data points with large feature values tend to appear earlier. The
-    ordered dataset is then split into `num_tasks` chunks.
-
-    The larger `exponent`, the more pronounced the sorting effect. Very large values will result in
-    (deterministic) sorting.
+    This scenario sorts the data according to a feature value (see `feature_idx`) and randomly
+    swaps data positions based on the setting of ``randomness`` (see `feature_idx`).
```
lballes, line 319: Remove one occurrence of "see `feature_idx`".

src/renate/benchmark/scenarios.py
```diff
@@ -244,39 +244,79 @@ def __init__(
...
+
+class _SortingScenario(Scenario):
+    """A scenario that _softly_ sorts a dataset by some score.
+
+    Randomness in the sorted order is induced by swapping the position of random pairs.
 
     Args:
         data_module: The source RenateDataModule for the the user data.
         num_tasks: The total number of expected tasks for experimentation.
+        randomness: A value between 0 and 1. 0 means no random swapping, 1 means random ordering.
```
lballes, line 270: Maybe be more specific in describing what this parameter does, which is introducing a certain percentage of random swaps. I don't think it is accurate to say that setting it to `1` results in a (completely) random ordering.

lballes:
We can merge this after fixing two small issues with the doc string.

Out of curiosity: Any particular reason you replaced the soft-sorting mechanism (`exponent`) with random swaps?

wistuba:
the visual ordering with very high exponent didn't match the sorting at all. furthermore, random swaps at least give the impression to better understand how the ordering is changed.

lballes (approved):

PR #159, opened 2023-03-01
Implementation of scalable DataBuffer

lballes:


src/renate/updaters/experimental/gdumb.py
```diff
@@ -56,9 +56,10 @@ def __init__(
 
     def load_state_dict(self, model: RenateModule, state_dict: Dict[str, Any], **kwargs) -> None:
         """Restores the state of the learner."""
+        if not hasattr(self, "_memory_buffer"):
+            self._memory_buffer = GreedyClassBalancingBuffer()
```
wistuba, line 60: is this happening because we create an instance without actually calling the constructor?

test/renate/updaters/experimental/test_er.py
```diff
@@ -81,11 +81,10 @@ def test_er_validation_buffer(tmpdir):
     )
     for i in range(3):
         for j in range(100):
-            assert torch.all(
-                torch.eq(
-                    datasets_val[i][j][0],
-                    model_updater._learner._val_memory_buffer[i * 100 + j][0][0],
-                )
+            assert torch.allclose(
```
wistuba, line 84: why is `allclose` needed? tolerance is pretty high

src/renate/memory/buffer.py
```diff
@@ -197,189 +136,204 @@ def set_transforms(
...
-        storage_mode: defaults.SUPPORTED_BUFFER_STORAGE_MODES = defaults.BUFFER_STORAGE_MODE,
-        transform: Optional[Callable] = None,
-        target_transform: Optional[Callable] = None,
-    ) -> None:
-        super().__init__(
-            storage_mode=storage_mode, transform=transform, target_transform=target_transform
-        )
-        self._max_size = 16
+    def save(self, target_dir: str) -> None:
+        if len(self):
```
wistuba, line 165: meaning of `len(self)` is not clear

src/renate/memory/buffer.py
```diff
@@ -161,34 +84,50 @@ def __getitem__(self, idx: int) -> Tuple[NestedTensors, DataDict]:
...
 
-    @abstractmethod
-    def _update(self, dataset: Dataset, metadata: DataDict) -> None:
-        pass
+    def set_metadata(self, key: str, values: torch.Tensor) -> None:
+        if values.size(0) != len(self):
+            raise ValueError()
+        if key not in self._metadata:
+            self._add_metadata_like({key: values})
+        self._metadata[key][: len(self)] = values.cpu()
```
wistuba, line 130: class docstring claims that devices are not changed

src/renate/memory/buffer.py
```diff
@@ -197,189 +136,204 @@ def set_transforms(
...
         self._max_size = state_dict["max_size"]
-        self._storage_mode = state_dict["storage_mode"]
         self._seed = state_dict["seed"]
         self._count = state_dict["count"]
-        self._size = state_dict["size"]
-        self._data_points = state_dict["data_points"]
-        self.metadata = state_dict["metadata"]
-
+        self._indices = state_dict["indices"]
+        self._data_point_prototype = state_dict["data_point_prototype"]
```
wistuba, line 160: what if storing data is not allowed?

src/renate/memory/buffer.py
```diff
@@ -197,189 +136,204 @@ def set_transforms(
...
-                    ).item()
-                ]
-                self[memory_idx_to_replace] = (
-                    dataset[i],
-                    {key: value[i] for key, value in metadata.items()},
-                )
-                self._class_to_index_map[largest_class].remove(memory_idx_to_replace)
-                self._record_class_index(dataset[i], memory_idx_to_replace)
-
+                largest_class, max_length = self._get_largest_class()
```
wistuba, line 308: not very efficient (compute and code). one could keep a counter around (array mapping class index to count) and then use argmax.

src/renate/updaters/experimental/gdumb.py
lballes: Yes, as you can see in the old version, we already had to create a new buffer object in `load_state_dict` for precisely that reason. The `if` clause makes sure this only happens when we re-create the learner, not when we're just reloading an old state. (That's necessary because the new buffer implementation relies on persistence of its internal state when `load_state_dict` is called without `load`.)

src/renate/memory/buffer.py
```diff
@@ -104,53 +33,47 @@ class DataBuffer(Dataset, ABC):
...
-
-        self._data_points = None
-        self.metadata = None
-        self._size = 0
-
-        self._transform = transform
-        self._target_transform = target_transform
+        self._datasets = []
+        self._indices = {}
+        self._data_point_prototype = None
```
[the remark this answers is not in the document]
lballes: We  keep around a prototypical data point to infer shape and data type. That is because the memory-mapped tensors that we use for storage don't save that information. The prototype just becomes part of the buffer's state dict, so we can re-infer this info.

src/renate/memory/buffer.py
lballes: I guess that passage of the doc string is obsolete. Technically the buffer is now agnostic to the device placement of the data (not the metadata). Metadata is always kept on CPU, so it makes sense to ensure this here so that people can input metadata values (e.g., logits) that reside on GPU.

src/renate/memory/buffer.py
lballes: Agreed, but let's fix separately, okay?

wistuba (approved):

PR #160, opened 2023-03-02
Update instructions to avoid users putting account-id in code.

geoalgo:
*Description of changes:* avoid encouraging users to put account-id in code to avoid their id to be leaked by accident

### DIFF
--- a/examples/simple_classifier_cifar10/start_with_hpo.py
+++ b/examples/simple_classifier_cifar10/start_with_hpo.py
@@ -1,6 +1,6 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
-
+import boto3
 from syne_tune.backend.sagemaker_backend.sagemaker_utils import get_execution_role
 from syne_tune.config_space import choice, loguniform, uniform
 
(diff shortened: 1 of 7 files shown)


examples/simple_classifier_cifar10/start_with_hpo.py
```diff
@@ -20,8 +20,7 @@
 }
 
 if __name__ == "__main__":
-
-    AWS_ID = "0123456789"  # use your AWS account id here
+    AWS_ID = boto3.client("sts").get_caller_identity().get("Account")
     AWS_REGION = "us-west-2"  # use your AWS preferred region here
```
hfurkanbozkurt, line 24: It would be great to change region as well:
```
AWS_REGION = boto3.session.Session().region_name
```

geoalgo: This is subjective and not as critical I believe, I think it is best done by the core contributor if they prefer this other way.

wistuba: I'm fine with it @hfurkanbozkurt @geoalgo feel free to add it if you want to

geoalgo: Personally, I think having the region in code is fine for this example and may even be better as it shows how you can change the region from the default one easily (which you need to do if you want to have some instances ATM, for instance g5 are only in us-east).

wistuba (approved):

PR #162, opened 2023-03-10
Custom Function Arguments Allow For More Types

wistuba:
Previously, custom arguments had to be of type `str` which required conversion. This PR will allow using types `int`, `float`, `bool`, `tuple`, and `list`.

Documentation was updated to reflect our changes made for custom function arguments.

Addresses #115 

### DIFF
--- a/src/renate/cli/parsing_functions.py
+++ b/src/renate/cli/parsing_functions.py
@@ -135,6 +135,19 @@ def get_updater_and_learner_kwargs(
     return updater_class, learner_kwargs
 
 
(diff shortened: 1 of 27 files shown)


src/renate/cli/parsing_functions.py
```diff
@@ -134,10 +135,35 @@ def get_updater_and_learner_kwargs(
     return updater_class, learner_kwargs
 
 
+def _cast_arguments(arguments: Dict[str, Dict[str, Any]], args_dict: Dict[str, str]) -> None:
```
lballes, line 138: Naming of the function arguments is confusing, and so is the doc string. Can we improve that?

lballes (approved):

lballes (approved):

Issue #114, opened 2023-02-06
Integration Tests


wistuba:
Create integration tests and run them regularly to ensure high quality of Renate.

PR #163, opened 2023-03-10
Add integ test and end-to-end test workflows

wesk:
*Issue #, if available:* #114

*Description of changes:*

Add initial workflows for:
- integ tests
- end-to-end tests


examples/benchmarking/temp_donotmerge.py
```diff
@@ -0,0 +1,40 @@
+# before merging, pull @marwistu's updated version and use that instead of this file
```
wesk, line 1: TODO before merging

wistuba:
I'm currently adding a way to define integration tests: #177 
For step 1 we need to run run_quick_test.py for each file in suites/quick and pass the current commit id.

PR #166, opened 2023-03-14
Bump Syne Tune

wistuba:
Bumps syne tune, allows to customize schedulers for experimentation and uses max_resource_attr instead of max_t.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ sagemaker>=2.112.0, <2.133.1
 syne-tune[aws,gpsearchers]==0.4.0
 pytorch-lightning~=1.8.6
 Pillow>=9.0, <9.3.1
+tabulate>=0.9.0, <0.9.1
 torchmetrics~=0.10.3
 torchvision>=0.13.0, <0.14.2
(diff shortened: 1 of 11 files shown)


requirements.txt
```diff
@@ -4,9 +4,10 @@ pandas>=1.4.0, <1.5.3
 boto3>=1.26.0, <1.26.42
 requests>=2.28.0, <2.28.2
 sagemaker>=2.112.0, <2.133.1
-syne-tune[aws,gpsearchers,kde,raytune]~=0.3.3
+syne-tune[aws,gpsearchers]==0.4.0
 pytorch-lightning~=1.8.6
 Pillow>=9.0, <9.3.1
+tabulate>=0.9.0, <0.9.1
```
lballes, line 10: Where does this requirement come from?

lballes (approved):

requirements.txt
wistuba: we use tabulate for experimentation to pretty print. earlier it was installed with syne tune. we removed those dependencies and now we need to list them individually. might be a good idea to eventually remove it

Issue #107, opened 2023-01-27
Example config file for nlp datasets


mridulgarg11:
Hi authors, is there any example of a config file for nlp datasets, especially around applying transformations? I’m trying to create a minimal example. Here's the config file I'm using-

```python
# renate_config.py
from pathlib import Path
from typing import Callable, Optional, Union

import torch
import torchtext.transforms as T
from torch.hub import load_state_dict_from_url
from torchvision.transforms import Lambda

from renate.benchmark.models.mlp import MultiLayerPerceptron

from renate.models import RenateModule

from renate import defaults
from renate.benchmark.datasets.nlp_datasets import TorchTextDataModule
from renate.benchmark.scenarios import ClassIncrementalScenario, Scenario


def data_module_fn(data_path: Union[Path, str], chunk_id: int, seed: int = defaults.SEED) -> Scenario:
    """Returns a class-incremental scenario instance.
    The transformations passed to prepare the input data are required to convert the data to
    PyTorch tensors.
    """
    data_module = TorchTextDataModule(
        str(data_path),
        dataset_name="AG_news",
        val_size=0.1,
        seed=seed,
    )

    class_incremental_scenario = ClassIncrementalScenario(
        data_module=data_module,
        class_groupings=[[0, 1], [2, 3]],
        chunk_id=chunk_id,
    )
    return class_incremental_scenario


def model_fn(model_state_url: Optional[Union[Path, str]] = None) -> RenateModule:
    """Returns a model instance."""
    if model_state_url is None:
        model = MultiLayerPerceptron(
            num_inputs=256, num_outputs=4, num_hidden_layers=2, hidden_size=64
        )
    else:
        state_dict = torch.load(str(model_state_url))
        model = MultiLayerPerceptron.from_state_dict(state_dict)
    return model


def train_transform() -> Callable:
    """Returns a transform function to be used in the training."""
    padding_idx = 1
    bos_idx = 0
    eos_idx = 2
    max_seq_len = 256
    xlmr_vocab_path = r"https://download.pytorch.org/models/text/xlmr.vocab.pt"
    xlmr_spm_model_path = r"https://download.pytorch.org/models/text/xlmr.sentencepiece.bpe.model"
    text_transform = T.Sequential(
        T.SentencePieceTokenizer(xlmr_spm_model_path),
        T.VocabTransform(load_state_dict_from_url(xlmr_vocab_path)),
        T.Truncate(max_seq_len - 2),
[...]

PR #171, opened 2023-03-23
Add example for fine-tuning of a transformer model 

lballes:
Resolves issue #107 


examples/nlp_finetuning/start_with_hpo.py
```diff
@@ -0,0 +1,53 @@
...
+    run_training_job(
+        config_space=config_space,
+        mode="max",
+        metric="val_accuracy",
+        updater="ER",  # we train with Experience Replay
+        max_epochs=5,
+        # we select the first chunk of our dataset, you will probably not need this in practice
+        chunk_id=0,
+        config_file="renate_config.py",
+        requirements_file="../../requirements.txt",
```
610v4nn1, line 36: This works only as far as the example is in the right folder. 
You can create an ad-hoc requirements.txt in this folder such that it will be tied to the example.

examples/nlp_finetuning/start_without_hpo.py
```diff
@@ -0,0 +1,47 @@
...
+        # overwriting
+        output_state_url=f"s3://sagemaker-{AWS_REGION}-{AWS_ID}/renate-training-nlp-finetuning/",
+        # uncomment the line below only if you already created a model with this script and you want
+        # to update it
+        # state_url=f"s3://sagemaker-{AWS_REGION}-{AWS_ID}/renate-training-nlp-finetuning/",
+        backend="sagemaker",  # run on SageMaker, select "local" to run this locally
+        role=get_execution_role(),
+        instance_count=1,
+        instance_type="ml.g4dn.xlarge",
+        job_name="renate-training-nlp-finetuning",
```
610v4nn1, line 46: Similar comments as above

src/renate/training/training.py
```diff
@@ -515,7 +515,7 @@ def _execute_training_and_tuning_job_locally(
     config_space["updater"] = updater
     config_space["max_epochs"] = max_epochs
     config_space["config_file"] = config_file
-    config_space["prepare_data"] = False
+    config_space["prepare_data"] = True
```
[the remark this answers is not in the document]
wistuba: nice find. this was probably introduced with my latest change. you could try `"False"` instead and maybe that works. Otherwise, I would need to dive deeper into this. It should not be True since otherwise we keep downloading data for each new training job.

Background: Previously this value was set to 0 and we used argparse to read it as an int. Now, this is considered an argsparse str and we should correctly convert it to a bool. Since this is not happening, it might be that 1) input is wrong or 2) that there might be a deeper problem that will cause problems with all bool inputs. My guess right now is option 3: something is going wrong pre-cli logic.

examples/nlp_finetuning/renate_config.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def train_transform() -> Callable:
+    """Returns a transform function to be used in the training."""
+    return None
+
+
+def test_transform() -> Callable:
+    """Returns a transform function to be used for validation or testing."""
+    return None
```
wistuba, line 52: i think we can drop them

examples/nlp_finetuning/renate_config.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+    model = RenateWrapper(transformer_model, loss_fn=torch.nn.CrossEntropyLoss())
+    if model_state_url is not None:
+        state_dict = torch.load(str(model_state_url))
+        model.load_state_dict(state_dict)
+    return model
+
+
+def data_module_fn(
+    data_path: Union[Path, str], chunk_id: int, seed: int = defaults.SEED
```
wistuba, line 29: replace `chunk_id: int` with `dataset_name: str` and enjoy the magic :)

examples/nlp_finetuning/start_with_hpo.py
```diff
@@ -0,0 +1,53 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
wistuba, line 1: do we want to have each example with both hpo and without? seems a bit too much for me

examples/nlp_finetuning/start_with_hpo.py
wistuba: do we need the requirements file? besides the current problem with the avalanche dependency, there is no need for it, right?

examples/nlp_finetuning/start_without_hpo.py
```diff
@@ -0,0 +1,47 @@
...
+    "memory_size": 300,
+    "loss_normalization": 0,
+    "loss_weight": 0.5,
+}
+
+if __name__ == "__main__":
+    AWS_ID = boto3.client("sts").get_caller_identity().get("Account")
+    AWS_REGION = "us-west-2"  # use your AWS preferred region here
+
+    run_training_job(
```
wistuba, line 26: do we ever use `chunk_id=1`? I like that you've used two datasets here. Why not use it as part of the example?

src/renate/training/training.py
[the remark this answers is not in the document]
wistuba: ok, I think it's much simpler than I've thought. there is a problem with the nlp dataset class. It seems to assume that prepare_data is always called before setup. however, the assumption is that prepare_data is only called once on the machine. based on my understanding following happens: 1) dataset object is created and prepare_data is called 2) another dataset object is created, only setup is called and this one is used for training. setting prepare_data to True will obviously "fix" this but I think we had a different understanding about the interface.

src/renate/updaters/experimental/er.py
```diff
@@ -206,7 +206,9 @@ def _update_memory_buffer(self, step_output: STEP_OUTPUT) -> None:
             metadata[
                 f"intermediate_representation_{i}"
             ] = intermediate_representation.detach().cpu()
-        dataset = Subset(self._current_train_dataset, step_output["train_data_idx"])
+        # Some datasets have problems using tensors as subset indices, convert to list of ints.
```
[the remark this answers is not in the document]
lballes: If `dataset` is a Huggingface dataset, `dataset[torch.tensor(1)]` fails. So I convert the subset indices to native `int`.

src/renate/training/training.py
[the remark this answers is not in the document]
lballes: You are absolutely right. I'll fix this. Thanks for catching this.

examples/nlp_finetuning/renate_config.py
```diff
@@ -0,0 +1,52 @@
...
+from renate.models import RenateModule
+from renate.models.renate_module import RenateWrapper
+
+
+def model_fn(model_state_url: Optional[Union[Path, str]] = None) -> RenateModule:
+    """Returns a DistilBert classification model."""
+    transformer_model = transformers.DistilBertForSequenceClassification.from_pretrained(
+        "distilbert-base-uncased", num_labels=2, return_dict=False
+    )
+    model = RenateWrapper(transformer_model, loss_fn=torch.nn.CrossEntropyLoss())
```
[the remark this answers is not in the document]
lballes: Yes, we can try that. I'll open an issue.

examples/nlp_finetuning/start_with_hpo.py
lballes: We need to make sure that all the dependencies are installed, right? Previously, we did this by adding a `requirements.txt` with `renate` listed. But that doesn't work if you're on `dev` and dependencies have changed compared to the PyPI version. I can copy the `requirements.txt` to the example folder, but then we need to make sure it stays in sync.

examples/nlp_finetuning/start_with_hpo.py
```diff
@@ -0,0 +1,53 @@
...
+        role=get_execution_role(),
+        instance_count=1,
+        instance_type="ml.g4dn.xlarge",
+        max_num_trials_finished=20,
+        scheduler="asha",  # run ASHA to optimize our hyperparameters
+        # if you use a big instance with multiple GPUs you can multiple workers evaluating
+        # configuration in parallel
+        # n_workers=4,
+        job_name="renate-training-nlp-finetuning",
+    )
```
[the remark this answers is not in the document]
lballes: All the examples only train on a single chunk, right? Or am I missing something?

examples/nlp_finetuning/start_with_hpo.py
```diff
@@ -0,0 +1,53 @@
...
+        # we select the first chunk of our dataset, you will probably not need this in practice
+        chunk_id=0,
+        config_file="renate_config.py",
+        requirements_file="../../requirements.txt",
+        # replace the url below with a different one if you already ran it and you want to avoid
+        # overwriting
+        output_state_url=f"s3://sagemaker-{AWS_REGION}-{AWS_ID}/renate-training-nlp-finetuning/",
+        # uncomment the line below only if you already created a model with this script and you want
+        # to update it
+        # state_url=f"s3://sagemaker-{AWS_REGION}-{AWS_ID}/renate-training-nlp-finetuning/",
```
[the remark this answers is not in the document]
lballes: Yes, thanks.

examples/nlp_finetuning/renate_config.py
lballes: Will do.

examples/nlp_finetuning/start_with_hpo.py
lballes: Yes, I'm happy to have this one without HPO only.

examples/nlp_finetuning/start_without_hpo.py
lballes: None of the examples actually perform more than one update, right? We would need to wait for the first run to finish, before starting the second one.

examples/nlp_finetuning/renate_config.py
lballes: Right, forgot about that.

examples/nlp_finetuning/start_without_hpo.py
wistuba: true, in this case we are using sagemaker. if we want to keep it that way, my question is why the config file has the option to use a second dataset.

lballes:
I changed back to using `chunk_id` to switch between the two datasets. I think it is a bit unintuitive for an example, when the `dataset_name` is part of the `config_space`. We explicitly introduced this `chunk_id` parameter in the `run_training_job` function to switch between chunks in the examples, so I think we should keep it consistent.

PR #177, opened 2023-03-28
Framework to Run Integration Tests

wistuba:
Updater, dataset etc. settings are defined via a json.
A test itself is also defined as a json.
A test is run via `run_quick_test.py --test-file er.json`. Results are stored on S3. Current tests are assumed to run locally.


610v4nn1:
I don't mind having a few configuration files stored for the integration tests.
At the same time, I wonder if we could not just generate a dictionary of dictionaries on-the-fly (possibly saving that in a json) and have a compact piece of code which is easier to understand and maintain.

wistuba:
I would like to keep the files in `suites` separate since it will allow us to iterate over them and start a test job for each with github actions. for the others you suggest to merge them into a single json file?

610v4nn1 (approved):

PR #179, opened 2023-03-30
Augment checkpointing callback to ensure that `on_model_update_end` is called

lballes:

### DIFF
--- a/src/renate/updaters/experimental/er.py
+++ b/src/renate/updaters/experimental/er.py
@@ -11,7 +11,7 @@ from pytorch_lightning.utilities.types import STEP_OUTPUT
 from torch.utils.data import DataLoader, Dataset, Subset
 
 from renate import defaults
-from renate.data.datasets import _EnumeratedDataset
+from renate.data.datasets import _EnumeratedDataset, _TransformedDataset
 from renate.memory.buffer import DataDict
 from renate.models import RenateModule
 from renate.types import NestedTensors
@@ -80,21 +80,26 @@ class BaseExperienceReplayLearner(ReplayLearner, abc.ABC):
 
     def on_model_update_start(
         self, train_dataset: Dataset, val_dataset: Dataset, task_id: Optional[str] = None
-    ) -> Tuple[DataLoader, DataLoader]:
+    ) -> None:
         """Called before a model update starts."""
+        super().on_model_update_start(train_dataset, val_dataset, task_id)
         self._set_memory_loader()
-        self._current_train_dataset = train_dataset
-        train_loader, val_loader = super().on_model_update_start(
-            train_dataset, val_dataset, task_id
+
+    def train_dataloader(self) -> DataLoader:
+        train_dataset = _EnumeratedDataset(
+            _TransformedDataset(
+                self._train_dataset,
+                transform=self._train_transform,
+                target_transform=self._train_target_transform,
+            )
         )
-        train_loader = DataLoader(
-            _EnumeratedDataset(train_loader.dataset),
+        return DataLoader(
+            train_dataset,
             batch_size=self._batch_size,
             shuffle=True,
             generator=self._rng,
             pin_memory=True,
         )
-        return train_loader, val_loader
 
     def on_train_start(self) -> None:
         """PyTorch Lightning function to be run at the start of the training."""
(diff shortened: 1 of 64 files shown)


src/renate/updaters/experimental/er.py
```diff
@@ -80,21 +80,26 @@ def _create_metrics_collections(
...
-    ) -> Tuple[DataLoader, DataLoader]:
+    ) -> None:
         """Called before a model update starts."""
+        super().on_model_update_start(train_dataset, val_dataset, task_id)
         self._set_memory_loader()
-        self._current_train_dataset = train_dataset
-        train_loader, val_loader = super().on_model_update_start(
-            train_dataset, val_dataset, task_id
+
+    def train_dataloader(self) -> DataLoader:
```
610v4nn1, line 88: Maybe I am missing something, but why cannot you use the one in the Learner base class?

src/renate/updaters/learner.py
```diff
@@ -94,7 +94,7 @@ def __init__(
 
     def _post_init(self) -> None:
         self._rng = get_generator(self._seed)
-        self._val_enabled = False
+        # self._val_enabled = False
```
610v4nn1, line 97: nit: remove if it's not needed

src/renate/updaters/model_updater.py
```diff
@@ -277,34 +298,28 @@ def _load_learner(
...
-            if val_loader is not None:
-                callbacks.insert(
-                    0,
-                    EarlyStopping(
-                        monitor=self._metric,
-                        mode=self._mode,
-                    ),
-                )
+            if learner._val_enabled:
+                print("Using early stopping!!!!")
```
610v4nn1, line 319: nit: remove print and use a warning if you want to communicate something important

src/renate/updaters/model_updater.py
```diff
@@ -277,34 +298,28 @@ def _load_learner(
...
+                callbacks.insert(0, EarlyStopping(monitor=self._metric, mode=self._mode))
             else:
+                print("!!! NOT using early stopping!!!!")
```
610v4nn1, line 322: nit: there is always a warning for this

src/renate/updaters/experimental/er.py
lballes: We add the enumeration here, which is used for the memory updates after each step.

src/renate/updaters/learner.py
lballes: Sorry, left over from debugging.

src/renate/updaters/model_updater.py
lballes: Sorry, left over from debugging.

src/renate/updaters/model_updater.py
lballes: Sorry, left over from debugging.

wistuba:
You can also add a quick integration test for offline er.

src/renate/updaters/model_updater.py
```diff
@@ -277,33 +298,25 @@ def _load_learner(
...
         self,
         learner: Learner,
-        train_loader: DataLoader,
-        val_loader: DataLoader,
         use_syne_tune_callback: bool = True,
     ) -> None:
         callbacks: List[Callback] = []
         if use_syne_tune_callback:
-            callbacks.append(SyneTuneCallback(val_loader is not None))
+            callbacks.append(SyneTuneCallback(learner._val_enabled))
```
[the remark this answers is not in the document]
lballes: I guess this info isn't directly available to the model updater. I can make it public `learner.val_enabled` or a method `learner.val_enabled()`?

test/renate/models/test_renate_module.py
```diff
@@ -198,11 +198,11 @@ def test_renate_model_multihead_save_and_load(tmpdir, test_case, test_cls):
 def test_renate_model_add_params(test_case):
     model, train_dataset, _ = test_case
     model_updater = pytest.helpers.get_simple_updater(model)
-    model = model_updater.update(train_dataset=train_dataset, task_id="second_task")
+    model_updater.update(train_dataset=train_dataset, task_id="second_task")
     assert model._tasks_params_ids == {TASK_ID, "second_task"}
```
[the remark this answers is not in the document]
lballes: We previously had `Learner.on_model_update_end` return the final model (which did nothing but return `Learner._model`). Now `on_model_update_end` is no longer called explicitly by us, but by the checkpointing callback. That's why I removed the return arg and just relied on the assumption that the model is updated in-place.

We could have `model_updater.update` return the `model`, but this would be pretty much exactly the same, since it could only return `ModelUpdater._model`.

Not sure what the best solution is here.

src/renate/updaters/model_updater.py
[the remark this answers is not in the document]
wistuba: is something like `len(learner.validation_loader()` possible or might that be too computational expensive?

wistuba, line 305: did you forget to update this?

wistuba (changes requested):

wistuba (approved):

PR #187, opened 2023-04-04
Make Wild Time Available in Benchmarking

wistuba:
Added some time-incremental benchmarks. Coverage is poor since we cannot reliably load the wild time data as part of the tests.


pyproject.toml
```diff
@@ -20,9 +20,13 @@ dynamic = ["version", "readme", "dependencies"]
 avalanche = [
     "avalanche_lib==0.3.1",
 ]
+benchmark = [
+    "wild-time-data==0.0.4",
```
610v4nn1, line 24: I am bit puzzled by this dependency because I thought this was a way to get rid of the WildTime dependencies but we are still going to install sklearn etc.

Can we include in Renate the little code we need and rely on the serialized data to decouple it from WildTime?
I am happy to discuss this offline if you want since it's a more complex question.

src/renate/benchmark/datasets/wild_time_data.py
```diff
@@ -0,0 +1,90 @@
...
+    Args:
+        data_path: the path to the folder containing the dataset files.
+        dataset_name: Name of the wild time dataset.
+        src_bucket: the name of the s3 bucket. If not provided, downloads the data from original
+            source.
+        src_object_name: the folder path in the s3 bucket.
+        time_step: Time slice to be loaded.
+        tokenizer: Tokenizer to apply to the dataset. See https://huggingface.co/docs/tokenizers/
+            for more information on tokenizers.
+        tokenizer_kwargs: Keyword arguments passed when calling the tokenizer's ``__call__``
```
610v4nn1, line 27: It would be great to report which one was used in the paper describing the dataset

src/renate/benchmark/scenarios.py
```diff
@@ -378,3 +379,27 @@ def _get_scores(self, dataset: Dataset) -> List[float]:
             )
             scores.append(value[np.argmax(count)])
         return scores
+
+
+class WildTimeScenario(Scenario):
```
610v4nn1, line 384: In the paper there are at least two scenarios and we may want to create a couple of variants as well.
I think the name of this class is too generic at the moment.

I also don't see if the data is split in chunks as in the paper, but I guess that's part of the package we are taking a dependency on. Is that correct?

wistuba: I think we can change the name as soon as we have a second class so we know what a good name would be.

The data is already split by time, nothing to do here.

src/renate/benchmark/datasets/wild_time_data.py
wistuba: The use DistilBERT but not really anything specific for the dataset.

wistuba:
I've added a bit more explanation explaining the difference to the original scenario.

610v4nn1 (approved):

PR #188, opened 2023-04-06
add initial integ test workflows

wesk:
*Description of changes:*

Wire up the quick tests to run on GitHub Actions with each PR. 

Observed that the accuracy values differ between Mac and Linux 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,7 @@
+**/__pycache__/
+.DS_Store
+*.egg-info
+std.out
+
+renate_working_dir/
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,0 @@
-**/__pycache__/
-.DS_Store
-*.egg-info
-std.out
-tmp/
-
-renate_working_dir/
-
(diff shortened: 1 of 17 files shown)


test/integration_tests/run_quick_test.py
```diff
@@ -65,4 +65,9 @@
     else:
         accuracies = []
 
+    print("Expected Accuracy:", test_config["expected_accuracy"])
+    print("Actual Accuracy:", accuracies)
```
wistuba, line 69: only print in case of failure

.gitignore
```diff
@@ -0,0 +1,8 @@
+**/__pycache__/
```
wistuba, line 1: can we drop this file?

wistuba (commented):
I have only minor comments

.gitignore
wesk: Sure, I pulled it out into a separate PR: https://github.com/awslabs/Renate/pull/190

wistuba (approved):

PR #190, opened 2023-04-11
Create `.gitignore` file

wesk:
*Issue #, if available:*

*Description of changes:*

Create a `.gitignore` file for the repo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+**/__pycache__/
+.DS_Store
+*.egg-info
+std.out
+tmp/
+
+renate_working_dir/
+
+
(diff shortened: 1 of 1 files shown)


wistuba (changes requested):
Would be great to cover a few more files and structure it a bit better. Check https://github.com/github/gitignore/blob/main/Python.gitignore for an example

wesk:
> Would be great to cover a few more files and structure it a bit better. Check https://github.com/github/gitignore/blob/main/Python.gitignore for an example

Updated to use the template

wistuba:
Thanks for your contribution. We decided to use our own .gitignore files instead of a shared one.

PR #195, opened 2023-04-18
Fix Doc Building

### LINKED ISSUE #196: Dev Documentation not up-to-date
I noticed that our dev documentation is not up-to-date. It appears, that since the avalanche integration building no longer works. Our test failed to spot it since it was installing dependencies differently to how RTD is doing it.

I'm working on rectifying the test and fixing all issues we've added over the time as part of this PR: #195

wistuba:
Our doc building action installs Renate[dev] while rtd does not. Therefore, our tests are passing but the docs are not updating anymore. This PR fixes the github action to reflect what rtd is doing and a hotfix addresses the problem.

Closes #196 

### DIFF
--- a/src/renate/updaters/avalanche/plugins.py
+++ b/src/renate/updaters/avalanche/plugins.py
@@ -1,7 +1,7 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 from pathlib import Path
-from typing import Any, Callable, Dict, IO, Optional, Union
+from typing import Callable, Dict, IO, Optional, Union
 
 import torch
 from avalanche.training.plugins.checkpoint import CheckpointPlugin, FileSystemCheckpointStorage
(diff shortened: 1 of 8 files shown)


wesk:
For Syne Tune I just followed this: https://docs.readthedocs.io/en/stable/guides/git-integrations.html#provider-specific-instructions

No code-changes needed, wondering if something similar can be done for Renate

wistuba:
This is not about updating the files on rtd but on catching errors in the doc building process. The reason rtd is not updating the files because the building is failing.

src/renate/updaters/avalanche/plugins.py
```diff
@@ -19,11 +19,6 @@ def _make_checkpoint_file_path(self, checkpoint_name: str) -> Path:
     def checkpoint_exists(self, checkpoint_name: str) -> bool:
         return self._make_checkpoint_file_path(checkpoint_name).exists()
 
-    def load_checkpoint(self, checkpoint_name: str, checkpoint_loader) -> Any:
-        checkpoint_file = self._make_checkpoint_file_path(checkpoint_name)
-        with open(checkpoint_file, "rb") as f:
-            return checkpoint_loader(f)
-
```
610v4nn1, line 26: This change has nothing to do with sphinx, is it intended?

610v4nn1 (approved):

src/renate/updaters/avalanche/plugins.py
wistuba: not sure why but sphinx will complain about that code. it is also identical to the parent's code, so we can drop it: https://github.com/ContinualAI/avalanche/blob/v0.3.1/avalanche/training/plugins/checkpoint.py#L335

PR #213, opened 2023-05-03
Add NLP Components to Benchmarking

wistuba:
Add RenateModule for Hugging Face transformer and make datasets and transformers available in benchmarking script. Tests are failing for other reasons :(

### DIFF
--- a/examples/nlp_finetuning/renate_config.py
+++ b/examples/nlp_finetuning/renate_config.py
@@ -1,7 +1,6 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
-from pathlib import Path
-from typing import Optional, Union
+from typing import Optional
 
 import torch
 import transformers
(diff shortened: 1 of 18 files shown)


examples/nlp_finetuning/renate_config.py
```diff
@@ -13,26 +12,24 @@
 from renate.models.renate_module import RenateWrapper
 
 
-def model_fn(model_state_url: Optional[Union[Path, str]] = None) -> RenateModule:
+def model_fn(model_state_url: Optional[str] = None) -> RenateModule:
```
lballes, line 15: If we're disallowing path objects here, should we also do that in the other examples?

src/renate/benchmark/models/transformer.py
```diff
@@ -0,0 +1,43 @@
...
+
+import torch
+import torch.nn as nn
+from torch import Tensor
+from transformers import AutoModelForSequenceClassification
+
+from renate.models import RenateModule
+
+
+class HuggingFaceSequenceClassificationTransformer(RenateModule):
```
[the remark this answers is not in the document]
wistuba: the RenateWrapper does not allow to save any additional arguments as in this case `pretrained_model_name` and `num_outputs`.

test/integration_tests/run_quick_test.py
```diff
@@ -67,4 +67,4 @@
 
     # Noticed different accuracy scores across Mac and GitHub Actions Workflows (which run on Linux)
     # TODO see if we can align the Mac and Linux results
-    assert pytest.approx(test_config["expected_accuracy"]) == accuracies
+    assert pytest.approx(test_config["expected_accuracy"]) == accuracies, accuracies
```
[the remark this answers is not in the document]
wistuba: if fails, prints `accuracies`

examples/nlp_finetuning/renate_config.py
wistuba: the use of complex types as Path is no longer supported for our *_fn functions since 0.2.0. I can check for more occurrences but I should have removed most of them before

src/renate/benchmark/models/transformer.py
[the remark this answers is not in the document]
lballes: Makes sense.

examples/nlp_finetuning/renate_config.py
lballes: I see occurrences of `Path` in most example config files. If you don't want to remove then now, we should create an issue to do it later.

lballes (dismissed):

examples/nlp_finetuning/renate_config.py
wistuba: I've updated the remaining config files as well as the documentation.

lballes (approved):

examples/nlp_finetuning/renate_config.py
lballes: Nice, thanks!

PR #217, opened 2023-05-04
Replace memory batch size with a fraction of the total batch size

lballes:


src/renate/updaters/experimental/er.py
```diff
@@ -95,7 +95,7 @@ def train_dataloader(self) -> DataLoader:
         )
         return DataLoader(
             train_dataset,
-            batch_size=self._batch_size,
```
alexanderbuchholz, line 98: Nit: Would it make sense to add a check for the value of `self._batch_memory_frac`? 
I.e., add as statement of the form 
`assert 0 <= self._batch_memory_frac <= 1.`

src/renate/updaters/experimental/er.py
```diff
@@ -192,7 +192,7 @@ def training_step(
 
     def _sample_from_buffer(self, device: torch.device) -> Optional[Tuple[NestedTensors, DataDict]]:
         """Function to sample from the buffer, if buffer is populated."""
-        if self._memory_loader is not None and len(self._memory_buffer) >= self._memory_batch_size:
```
alexanderbuchholz, line 195: Is changing this statement safe? This could create a different behavior,  no?

src/renate/updaters/experimental/gdumb.py
```diff
@@ -44,6 +44,7 @@ def __init__(
     ) -> None:
         super().__init__(
             memory_size=memory_size,
+            batch_memory_frac=-1.0,  # Dummy value
```
wistuba, line 47: why not 1?

alexanderbuchholz, line 47: Should this raise an error?

src/renate/updaters/experimental/repeated_distill.py
```diff
@@ -229,7 +229,7 @@ def __init__(
         super().__init__(
             model=model,
             memory_size=memory_size,
-            memory_batch_size=batch_size,
+            batch_memory_frac=-1.0,
```
wistuba, line 232: is it not using the memory?

src/renate/updaters/experimental/offline_er.py
```diff
@@ -66,17 +65,32 @@ def on_model_update_start(
         self._num_points_current_task = len(train_dataset)
 
     def train_dataloader(self) -> DataLoader:
-        train_loader = super().train_dataloader()
-        loaders = {"current_task": train_loader}
-        if len(self._memory_buffer) > self._memory_batch_size:
+        """Returns the dataloader for training the model."""
+        loaders = {}
+        memory_batch_size = int(self._batch_memory_frac * self._batch_size)
+        if len(self._memory_buffer) >= memory_batch_size:
```
alexanderbuchholz, line 71: Not directly related to this PR, but would it make sense to add a check somewhere that the buffer size must always be larger than the memory size? Otherwise the memory is always ignored.

examples/benchmarking/class_incremental_learning_cifar10_der.py
```diff
@@ -11,8 +11,8 @@
     "learning_rate": 0.03,
     "alpha": 0.2,
     "beta": 0.5,
-    "batch_size": 32,
-    "memory_batch_size": 32,
+    "batch_size": 64,
+    "batch_memory_frac": 0.5,
```
wistuba, line 15: I'd prefer full words over abbreviations

src/renate/updaters/learner.py
```diff
@@ -404,14 +402,14 @@ class ReplayLearner(Learner, abc.ABC):
...
-        memory_batch_size: int = defaults.BATCH_SIZE,
+        batch_memory_frac: float = defaults.BATCH_MEMORY_FRAC,
         buffer_transform: Optional[Callable] = None,
         buffer_target_transform: Optional[Callable] = None,
         seed: int = defaults.SEED,
         **kwargs,
     ) -> None:
         super().__init__(seed=seed, **kwargs)
-        self._memory_batch_size = min(memory_size, memory_batch_size)
+        self._batch_memory_frac = min(batch_memory_frac, memory_size / self._batch_size)
```
alexanderbuchholz, line 412: and here we could check that `memory_size > batch_memory_frac * self._batch_size`

test/conftest.py
```diff
@@ -113,7 +113,7 @@ def pytest_collection_modifyitems(config, items):
     },
     OfflineExperienceReplayLearner: {
         "memory_size": 30,
-        "memory_batch_size": 20,
+        "batch_memory_frac": 0.6,
```
alexanderbuchholz, line 116: What is the `batch_size` here? if the total batch size is 30 and the memory batch size is 20 this implies `batch_memory_frac = 0.666666`. But the values here would imply `batch_size = 33.33333`. (Or is it me doing the computation wrong here?). Thanks

test/integration_tests/configs/suites/quick/offline-er.json
```diff
@@ -5,6 +5,6 @@
     "dataset": "cifar10.json",
     "backend": "local",
     "job_name": "class-incremental-mlp-offline-er",
-    "expected_accuracy": [0.7089999914169312, 0.5120000243186951],
```
alexanderbuchholz, line 8: why is the expected accuracy changing? This should not be the case, no?

alexanderbuchholz (commented):
Thanks for looking into this! Please see my comments below.

test/conftest.py
wistuba: true, this should imply that memory_batch_size will be 18

test/integration_tests/configs/suites/quick/offline-er.json
lballes: The behavior of Offline-ER is changed slightly by this PR. In the first update, it will now use the total batch size exclusively for the current task data, since no memory is available yet. Hence, small changes are to be expected and I updated the numbers here.

test/conftest.py
lballes: I didn't pay close attention to the numbers, since the tests don't depend on exact memory sizes.

src/renate/updaters/experimental/er.py
lballes: Yes, good point. Will add this.

src/renate/updaters/experimental/er.py
```diff
@@ -192,7 +192,7 @@ def training_step(
...
+        if self._memory_loader is not None:
```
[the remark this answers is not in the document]
lballes: Will add a check.

src/renate/updaters/experimental/er.py
```diff
@@ -219,10 +219,11 @@ def _update_memory_buffer(self, step_output: STEP_OUTPUT) -> None:
 
     def _set_memory_loader(self) -> None:
         """Create a memory loader from a memory buffer."""
-        if self._memory_loader is None and len(self._memory_buffer) >= self._memory_batch_size:
+        memory_batch_size = int(self._batch_memory_frac * self._batch_size)
+        if self._memory_loader is None and len(self._memory_buffer) >= memory_batch_size:
```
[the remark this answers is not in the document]
lballes: Because this is creating the memory loader, so we want to make sure it will return at least one batch. In `_sample_memory_buffer` it should then suffice to check for the existence of a memory loader, so I removed the superfluous second part.

src/renate/updaters/experimental/gdumb.py
lballes: Will change that.

src/renate/updaters/experimental/repeated_distill.py
lballes: It is *only* sampling from the memory using the entire batch size. So passing a dummy value to the parent class (which probably means this inheritance structure is not ideal...)

src/renate/updaters/experimental/gdumb.py
lballes: Will add a check for to be in [0, 1] and then use 0.0 as a dummy value.

src/renate/updaters/learner.py
lballes: Yes, I'll add that.

src/renate/updaters/experimental/gdumb.py
```diff
@@ -45,9 +45,9 @@ def __init__(
         super().__init__(
             memory_size=memory_size,
             seed=seed,
+            # batch_memory_frac=0.0,
```
prabhuteja12, line 48: is this unused?

test/conftest.py
```diff
@@ -69,7 +69,7 @@ def pytest_collection_modifyitems(config, items):
 LEARNER_KWARGS = {
     ExperienceReplayLearner: {
         "memory_size": 30,
-        "memory_batch_size": 20,
+        "batch_memory_frac": 0.6,
```
prabhuteja12, line 72: The ratios dont match: .6 * 30 != 20?

lballes: Yes, will adjust this. But the tests don't depend on the exact numbers. All unit tests are passing. It's the integration tests that puzzle me.

src/renate/updaters/experimental/offline_er.py
```diff
@@ -77,12 +76,28 @@ def on_model_update_start(
...
+            transform=self._train_transform,
+            target_transform=self._train_target_transform,
+        )
+        loaders = {
+            "current_task": DataLoader(
+                train_dataset,
+                batch_size=self._batch_size - memory_batch_size,
+                shuffle=True,
+                generator=self._rng,
+                pin_memory=True,
```
wistuba, line 94: was this the PR where you wanted to add the drop_last argument?

src/renate/updaters/avalanche/learner.py
```diff
@@ -73,10 +73,11 @@ class AvalancheReplayLearner(ReplayLearner, AvalancheLoaderMixin):
     def create_avalanche_learner(
         self, plugins: List[SupervisedPlugin], **kwargs: Any
     ) -> BaseSGDTemplate:
+        memory_batch_size = round(self._batch_memory_frac * self._batch_size)
```
wistuba, line 76: you could simply reuse `self._memory_batch_size`

src/renate/updaters/model_updater.py
```diff
@@ -411,8 +411,8 @@ def _fit_learner(
 
         strategy = create_strategy(self._devices, self._strategy)
         # Finetuning-equivalent epochs.
-        num_batches = len(learner._train_dataset) // learner._batch_size
-        num_batches += min(len(learner._train_dataset) % learner._batch_size, 1)
+        num_batches = len(learner._train_dataset) // 256  # learner._batch_size
+        num_batches += min(len(learner._train_dataset) % 256, 1)  # learner._batch_size, 1)
```
wistuba, line 415: not sure why this doesn't work on github, but it works locally. the reason the tests are failing is that a pass over the data has now effectively fewer batches.

what's a bit concerning is that I saw that the validation loss has changed. are there any changes to how the validation buffer gets populated or do we drop the last batch of the validation buffer? funnily, that does not happen for each epoch.

PR #220, opened 2023-05-05
Expose `limit_train_batches` flag of Lightning Trainer

lballes:


src/renate/cli/run_experiment_with_scenario.py
```diff
@@ -102,6 +102,12 @@ def run(self):
             default=defaults.MAX_EPOCHS,
             help=f"Number of epochs trained at most. Default: {defaults.MAX_EPOCHS}",
         )
+        argument_group.add_argument(
+            "--limit_train_batches",
+            type=int,
```
wistuba, line 107: typing indicates float or int

wistuba (changes requested):
you probably need to make limit_train_batches a string and parse to int or float manually afterwards

PR #224, opened 2023-05-09
Update training.py

610v4nn1:
*Issue #, if available:*

*Description of changes:*


wistuba (changes requested):
this would mean that by default we run each epoch for only a single batch. I assume the intention was to set the default to 100% of batches with 1.0. i see two solutions here: either use floats only or implement the logic outlined in #224

610v4nn1:
I was just testing a couple of things. I don't want to merge these changes. Sorry for the confusion.

PR #237, opened 2023-05-22
Add MMD covariate shift detector

lballes:
- Shift detectors reside in `renate.shift`
- Add base class for shift detectors, including functionality for feature extraction
- Add MMD covariate shift detector as first method


src/renate/shift/mmd_detectors.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Args:
+        feature_extractor: A pytorch model used as feature extractor.
+        num_permutations: Number of permutations for permutation test.
+        batch_size: Batch size used to iterate over datasets.
+        num_preprocessing_workers: Number of workers used in data loaders.
+        device: Device to use for computations inside the detector.
+    """
+
+    def __init__(
```
610v4nn1, line 29: I think that exposing the choice of the kernel to be used could be beneficial in the long term

src/renate/shift/mmd_helpers.py
```diff
@@ -0,0 +1,124 @@
...
+    def __init__(self):
+        pass
+
+    def _check_inputs(self, X0: torch.Tensor, X1: torch.Tensor):
+        assert X0.dim() == X1.dim() == 2
+        assert X0.size(1) == X1.size(1)
+        assert X0.dtype is X1.dtype
+
+
+class RBFKernel(Kernel):
```
610v4nn1, line 20: the kernels are not directly tied to MMD and it would be better to have them in a separate file (e.g., kernels.py)

src/renate/shift/mmd_helpers.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @torch.no_grad()
+    def __call__(self, X0: torch.Tensor, X1: torch.Tensor):
+        self._check_inputs(X0, X1)
+        dists = torch.cdist(X0, X1)
+        lengthscale = self._lengthscale or torch.median(dists)
+        return torch.exp(-0.5 * dists**2 / lengthscale**2)
+
+
+def mmd_gram(
```
610v4nn1, line 44: It's not clear to me why this method (and the next) is not a private method of the MMD detector class. You can also have an MMD base class if you prefer not to have them tied to the covariate shift.

test/renate/shift/test_detector.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@pytest.mark.parametrize(
+    "dataset",
+    [
+        torch.utils.data.TensorDataset(torch.randn(20, 2)),
+        torch.utils.data.TensorDataset(torch.randn(20, 2), torch.randint(0, 10, size=(20,))),
+    ],
+)
+@pytest.mark.parametrize("feature_extractor", [torch.nn.Linear(2, 4)])
```
610v4nn1, line 16: can we check the output of the feature extractor? for example by using a model initialized ad-hoc (e.g., all-ones)

test/renate/shift/test_mmd_helpers.py
```diff
@@ -0,0 +1,39 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+import pytest
+import torch
+
+from renate.shift.mmd_helpers import RBFKernel
+
+
```
610v4nn1, line 8: There are no tests for `mmd` and `mmd_gram`

src/renate/shift/mmd_detectors.py
```diff
@@ -0,0 +1,45 @@
...
+    """A kernel maximum mean discrepancy (MMD) test.
+
+    This test was proposed by
+
+    [1] Gretton, A., et al. A kernel two-sample test. JMLR (2012).
+
+    We currently do not expose the choice of kernel. It defaults to an RBF kernel with a lengthscale
+    set via the median heuristic.
+
+    Args:
```
610v4nn1, line 21: The outputs of the score method are not explained in the base class and not explained here. It's going to be hard for users to know when they get back a generic score, when they get a p-value, etc.

src/renate/shift/detector.py
```diff
@@ -0,0 +1,105 @@
...
+from typing import Optional
+
+from renate.utils.pytorch import move_tensors_to_device
+
+
+class ShiftDetector:
+    """Base class for distribution shift detectors.
+
+    Args:
+        batch_size: Batch size used to iterate over datasets.
```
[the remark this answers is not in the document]
lballes: Will clarify.

src/renate/shift/mmd_detectors.py
lballes: Yes, I will expand the doc string.

src/renate/shift/mmd_detectors.py
lballes: Agree, but I'd like to move along with this quickly. I think we can easily integrate other kernels later.

test/renate/shift/test_detector.py
lballes: Will do.

test/renate/shift/test_mmd_helpers.py
lballes: They are touched via the test for the MMD detector, but I will add similar tests that directly invoke these functions.

src/renate/shift/mmd_detectors.py
610v4nn1: OK, please add a TODO

lballes: Added an issue.

610v4nn1 (approved):

PR #273, opened 2023-05-30
Fix Offline-ER bug and change loss functions

wistuba:
Loss functions should no longer be reduced.


src/renate/updaters/experimental/offline_er.py
```diff
@@ -96,16 +97,31 @@ def training_step(
         else:
             alpha = self._loss_weight_new_data
         inputs, targets = batch["current_task"]
+        device = inputs.device
+        batch_size_current = len(inputs)
+        batch_size_mem = 0
+        if "memory" in batch:
+            (inputs_mem, targets_mem), _ = batch["memory"]
+            batch_size_mem = len(inputs_mem)
```
prabhuteja12, line 105: why len? and not shape[0]

PR #274, opened 2023-06-01
Add example of using renate in your own script

lballes:


examples/functional_usage/train.py
```diff
@@ -0,0 +1,37 @@
...
+### Instantiate renate ModelUpdater and run fine-tuning.
+updater = FineTuningModelUpdater(
+    model,
+    loss_fn,
+    optimizer="Adam",
+    learning_rate=3e-4,
+    batch_size=32,
+    max_epochs=3,
+    input_state_folder=None,
+    output_state_folder="renate_output",
```
wistuba, line 35: any deepspeed params we could showcase?

examples/functional_usage/train.py
```diff
@@ -0,0 +1,37 @@
...
+)
+updater.update(data_module.train_data(), data_module.val_data())
```
wistuba, line 37: make clear which model to use after this line for, e.g., evaluation on test

lballes: Yes, done.

examples/functional_usage/train.py
lballes: We could, but I guess the functional usage would be more appropriate for quick experiments, debugging, etc. So something you might run on your laptop?

610v4nn1: I would say folks can use this for experimentation on SM Notebooks instances where GPUs are available.
Anyway, we can add a pointer to the documentation if we don't want to include multi-GPU training from the start.

doc/examples/functional_usage.rst
```diff
@@ -0,0 +1,39 @@
+Using Renate Functionally
```
610v4nn1, line 1: IMO the title is too generic and not very accessible.

Maybe we should think about something like "Using Renate components to train a model in a custom script" ?

doc/examples/index.rst
```diff
@@ -7,4 +7,5 @@ Examples
 
     train_mlp_locally
     train_classifier_sagemaker
-    nlp_finetuning
\ No newline at end of file
+    nlp_finetuning
+    functional_usage
```
610v4nn1, line 11: I see the same problem I mentioned for the title

doc/examples/functional_usage.rst
```diff
@@ -0,0 +1,39 @@
...
+hyperparameters. Once the model updater is created, we initiate the training by calling its
+:py:meth:`~renate.updaters.model_updater.ModelUpdater.update` method and passing training and
+(optionally) validation datasets.
+
+.. literalinclude:: ../../examples/functional_usage/train.py
+    :lines: 27-37
+
+Once the training is terminated, your model is ready to deploy. Here, we just save its weights for
+later use.
+
```
610v4nn1, line 37: Maybe we can link the pytorch documentation explaining how to load the pt model

examples/functional_usage/train.py
```diff
@@ -0,0 +1,41 @@
...
+from renate.benchmark.datasets.nlp_datasets import HuggingFaceTextDataModule
+from renate.models.renate_module import RenateWrapper
+from renate.updaters.experimental.fine_tuning import FineTuningModelUpdater
+
+
+### Create model.
+transformer_model = transformers.DistilBertForSequenceClassification.from_pretrained(
+    "distilbert-base-uncased", num_labels=2, return_dict=False
+)
+model = RenateWrapper(transformer_model)
```
610v4nn1, line 15: It's not strictly related to this PR but it would be nice to discuss the pros and cons of doing this operation directly in the model updater.

doc/examples/functional_usage.rst
```diff
@@ -0,0 +1,39 @@
...
+:py:func:`~renate.training.training.run_training_job` function. In this example, we demonstrate how
+to write your own training script and use renate in a functional way. This can be useful, e.g., for
+debugging new components.
+
+Here, we use Renate to fine-tune a pretrained Transformer model on a sequence classification
+dataset. First, we create the model and a loss function. Since this is a static model, we simply
+wrap it using the :py:class:`~renate.models.renate_module.RenateWrapper` class.
+
+.. literalinclude:: ../../examples/functional_usage/train.py
+    :lines: 12-16
```
610v4nn1, line 14: if reduction = none is required, you need to explain why

lballes: Done.

doc/examples/functional_usage.rst
lballes: Done.

doc/examples/index.rst
lballes: Renamed.

examples/functional_usage/train.py
lballes: I would keep this without multi-GPU. This is not what we're trying to illustrate here. I'm now mentioning in the doc page that the ModelUpdater retrieves these arguments.

610v4nn1 (approved):

PR #275, opened 2023-06-01
Run sagemaker tests from GitHub Actions

wesk:
Initial PR to run the full test suite from GitHub Actions

### DIFF
--- a/test/integration_tests/run_experiment.py
+++ b/test/integration_tests/run_experiment.py
@@ -3,6 +3,7 @@
 import argparse
 import json
 from pathlib import Path
+import os
 
 import boto3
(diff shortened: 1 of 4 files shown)


test/integration_tests/run_experiment.py
```diff
@@ -59,6 +59,12 @@ def load_config(scenario_file, model_file, updater_file, dataset_file):
         default=12 * 3600,
         help="Maximum execution time.",
     )
+    parser.add_argument(
```
wistuba, line 62: in both files, this argument is considered optional but if not provided, the script will fail

wesk: Good catch, updated in next version

wistuba (approved):

PR #284, opened 2023-06-06
Patch Release 0.3.1

### LINKED ISSUE #283: moduleNotFoundError: No module named 'wild_time_data'
after installing renate via pip install renate on a new environment python 3.8,
when importing from renate.benchmark.scenarios import ClassIncrementalScenario
there is this in the title mentioned error.

Had to use pip install wild-time-data to mitigate the problem

wistuba:
Closes #283 


prabhuteja12 (approved):
Ideally we depend on tensorboard but lightning seems to work with either.

PR #285, opened 2023-06-06
Sync dev with main

wistuba:
Sync dev with Main


requirements.txt
```diff
@@ -8,8 +8,9 @@ syne-tune[aws,gpsearchers]==0.6.0
 pytorch-lightning>=1.8.0, <1.9.5
 Pillow>=9.0, <9.5.1
 tabulate>=0.9.0, <0.9.1
+tensorboardX>=2.5.0, <2.5.2
```
610v4nn1, line 11: why do we need to take a dependency on tensorboardX?

610v4nn1 (approved):

PR #289, opened 2023-06-08
Flag to remove intermediate tasks' states

prabhuteja12:
Large datasets' buffers can be too big and result in storage issues (noticed while training CLEAR). This flag eliminates backing up intermediate results and overwrites the same folder/files in each update. 

### DIFF
--- a/src/renate/benchmark/experimentation.py
+++ b/src/renate/benchmark/experimentation.py
@@ -384,6 +384,10 @@ def _execute_experiment_job_locally(
     cumulative_metrics = create_cumulative_metrics("classification")
     df = cumulative_metrics_summary(results, cumulative_metrics, num_updates - 1)
     save_pandas_df_to_csv(df, defaults.metric_summary_file(logs_url))
+    if not retain_intermediate_state:
+        move_to_uri(
+            defaults.hpo_file(working_directory), defaults.logs_folder(experiment_outputs_url)
+        )
     logger.info("### Cumulative results: ###")
     logger.info(df)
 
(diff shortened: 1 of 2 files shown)


src/renate/benchmark/experimentation.py
```diff
@@ -179,6 +180,10 @@ def execute_experiment_job(
         deterministic_trainer: When true the Trainer adopts a deterministic behaviour also on GPU.
             In this function this parameter is set to True by default.
         job_name: Name of the experiment job.
+        strategy: String denoting lightning distributed strategy.
+        precision: String for which precision to use.
+        retain_intermediate_state: Flag to retain intermediate models and buffer states.
```
wistuba, line 185: maybe rephrase to "retain states after each update"? otherwise very redundant description and not helpful for anyone who doesn't understand what "intermediate" refers to.

src/renate/defaults.py
```diff
@@ -57,6 +57,7 @@
 JOB_NAME = "renate"
 SUPPORTED_TUNING_MODE = ["min", "max"]
 SUPPORTED_TUNING_MODE_TYPE = Literal["min", "max"]
+RETAIN_INTERMEDIATE_STATE = True
```
wistuba, line 60: Let's set it to `False`.

wistuba (changes requested):

src/renate/defaults.py
prabhuteja12: The current mode of operation is `True`. We shouldn't alter that, right?

wistuba: given that we never used the intermediate states, I'm fine changing the default behavior

wistuba (approved):

PR #301, opened 2023-06-14
Introduce `RenateLightningModule`

wistuba:
Split Learner into two classes. RenateLightningModule contains the basic logic while Learner adds all CL-related stuff such as val_buffer


examples/simple_classifier_cifar10/renate_config.py
```diff
@@ -36,7 +37,7 @@ def data_module_fn(data_path: str, chunk_id: int, seed: int = defaults.SEED) ->
     )
     class_incremental_scenario = ClassIncrementalScenario(
         data_module=data_module,
-        class_groupings=[[0, 1, 2, 3, 4], [5, 6, 7, 8, 9]],
+        class_groupings=((0, 1, 2, 3, 4), (5, 6, 7, 8, 9)),
```
prabhuteja12, line 40: Why has this been changed?

src/renate/benchmark/experiment_config.py
```diff
@@ -317,7 +318,7 @@ def train_transform(dataset_name: str) -> Optional[transforms.Compose]:
     raise ValueError(f"Unknown dataset `{dataset_name}`.")
 
 
-def test_transform(dataset_name: str) -> Optional[transforms.Normalize]:
+def test_transform(dataset_name: str) -> Optional[Union[transforms.Normalize, transforms.Compose]]:
```
prabhuteja12, line 321: Can this be made a callable? This is very specific to existing transformations.

examples/nlp_finetuning/start.py
```diff
@@ -25,8 +25,8 @@
 
     run_training_job(
         config_space=config_space,
-        mode="max",
-        metric="val_accuracy",
+        mode="min",
```
prabhuteja12, line 28: Does this make a difference?

src/renate/updaters/learner.py
```diff
@@ -76,41 +64,35 @@ def __init__(
         self._learning_rate_scheduler = learning_rate_scheduler
         self._learning_rate_scheduler_interval = learning_rate_scheduler_interval
         self._batch_size = batch_size
-        self._train_transform = train_transform
-        self._train_target_transform = train_target_transform
-        self._test_transform = test_transform
-        self._test_target_transform = test_target_transform
         self._seed = seed
         self._task_id: str = defaults.TASK_ID
+        self._train_dataset = None
```
prabhuteja12, line 69: no typing :)

src/renate/updaters/learner.py
```diff
@@ -224,11 +182,17 @@ def forward(self, inputs: NestedTensors, task_id: Optional[str] = None) -> torch
             task_id = self._task_id
         return self._model(inputs, task_id=task_id)
 
+    def training_step_unpack_batch(
+        self, batch: Tuple[NestedTensors, torch.Tensor]
+    ) -> Tuple[NestedTensors, Any]:
+        inputs, targets = batch
```
prabhuteja12, line 188: This choice of what a batch comprises is specific to HF vs non-HF datasets? So, why is this specified in the base class?

src/renate/updaters/learner.py
```diff
@@ -253,9 +217,15 @@ def training_epoch_end(self, outputs: List[Union[Tensor, Dict[str, Any]]]) -> No
         if not self.val_enabled:
             self._log_metrics()
 
+    def validation_step_unpack_batch(
+        self, batch: Tuple[NestedTensors, torch.Tensor]
+    ) -> Tuple[NestedTensors, Any]:
```
prabhuteja12, line 222: same as above.

examples/simple_classifier_cifar10/renate_config.py
wistuba: To match typing.

examples/nlp_finetuning/start.py
wistuba: yes. accuracy is no longer computed by default and since it was not added to the renate config, using it wouldn't be possible. we can keep it but then we need to add the metric.

src/renate/benchmark/scenarios.py
```diff
@@ -56,7 +59,10 @@ def prepare_data(self) -> None:
     @abc.abstractmethod
```
prabhuteja12, line 59: What does an `abc` with code do?

src/renate/benchmark/experiment_config.py
prabhuteja12: Why is this also `Optional`?

prabhuteja12 (changes requested):
Few small changes. I agree with the overall structure of what is being accomplished.

src/renate/benchmark/experiment_config.py
wistuba: because it can return `None`

src/renate/benchmark/scenarios.py
wistuba: Forces the implementation of the function:

```
import abc

class A(abc.ABC):
    @abc.abstractmethod
    def foo(self):
        print("parent")
        
        
class B(A):
    def foo(self):
        super().foo()
        
class C(A):
    pass

A().foo()  # Fails
B().foo()  # Works
C().foo()  # Fails
```

prabhuteja12 (approved):

Issue #219, opened 2023-05-05
Replace ViT with Hugging Face ViT


wistuba:
We have our own vision transformer which might not work well with other libraries that build on top of Hugging Face.

prabhuteja12:
Fixed in #303 .

PR #303, opened 2023-06-16
Using HuggingFace ViT implementation (#219)

prabhuteja12:
Solves #219 

### DIFF
--- a/src/renate/benchmark/models/vision_transformer.py
+++ b/src/renate/benchmark/models/vision_transformer.py
@@ -1,14 +1,44 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
-from functools import partial
-from typing import Any, Callable, List, Optional
+from typing import Any, Optional, Tuple, Union
 
-import torch.nn as nn
-from torchvision.models.vision_transformer import ConvStemConfig, WeightsEnum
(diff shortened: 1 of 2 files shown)


src/renate/benchmark/models/vision_transformer.py
```diff
@@ -119,6 +144,7 @@ def __init__(self, **kwargs: Any) -> None:
             num_heads=12,
             hidden_dim=768,
             mlp_dim=3072,
+            pretrained_name="google/vit-base-patch16-224",
```
wistuba, line 147: why still pass the other arguments? they aren't used, right? or is this a requirement for reloading the model?

prabhuteja12: Point.

wistuba: Ok, I thought the idea was to reload it without `pretrained_name`. Now, it will not be able to reloaded since `pretrained_name` is missing? Can we maybe add a test for this model or do we already have that?

prabhuteja12: `model_fn` loads through `from_state_dict` if `model_state_url` is specified. So reloading (in principle) should be able to restore the weights.

prabhuteja12: What would a test for this do?

wistuba (approved):

PR #305, opened 2023-06-19
Fix Small Bug in Benchmarking Script and Add LR Scheduler to Experiment Config

wistuba:
- Benchmarking script fails if no intermediate states are used since it tries to move a file with a function that expects a folder
- Add LR scheduler to replicate CLEAR results

### DIFF
--- a/src/renate/benchmark/experiment_config.py
+++ b/src/renate/benchmark/experiment_config.py
@@ -1,9 +1,12 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
-from typing import List, Optional, Tuple, Union
+from functools import partial
+from typing import Callable, List, Optional, Tuple, Union
 
 import torch
 import wild_time_data
+from torch.optim import Optimizer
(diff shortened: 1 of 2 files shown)


src/renate/utils/file.py
```diff
@@ -41,21 +41,30 @@ def _parse_s3_url(s3_url: str) -> Tuple[str, str]:
...
+        src: Source directory or file.
+        dst: Target directory or file.
         ignore_extensions: List of extensions to ignore.
         copy: If `True`, copy instead of move.
     """
-    for src_dir, _, files in os.walk(source_dir):
-        dst_dir = src_dir.replace(source_dir, destination_dir, 1)
+    if Path(src).is_file():
+        Path(dst).mkdir(parents=True, exist_ok=True)
+        dst_file = os.path.join(dst, os.path.basename(src))
```
prabhuteja12, line 59: Why the mix of Pathlib and os.path?

prabhuteja12 (changes requested):
We could stick to Pathlib or the os.path instead of mixing them.

prabhuteja12 (approved):

PR #314, opened 2023-06-24
Upgrade to torchmetrics 0.11

wistuba:
Account for interface changes in torchmetrics 0.11.

### DIFF
--- a/examples/getting_started/renate_config.py
+++ b/examples/getting_started/renate_config.py
@@ -8,7 +8,7 @@ import torchvision
 from torch.nn import Parameter
 from torch.optim import AdamW, Optimizer
 from torch.optim.lr_scheduler import StepLR, _LRScheduler
-from torchmetrics import Accuracy
+from torchmetrics.classification import MulticlassAccuracy
 from torchvision.transforms import transforms
 
 from renate.data.data_module import RenateDataModule
@@ -93,7 +93,7 @@ def buffer_transform() -> Callable:
 
 
 def metrics_fn() -> Dict:
-    return {"accuracy": Accuracy()}
+    return {"accuracy": MulticlassAccuracy(num_classes=10)}
 
 
 def loss_fn() -> torch.nn.Module:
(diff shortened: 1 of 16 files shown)


src/renate/benchmark/experiment_config.py
```diff
@@ -361,5 +359,6 @@ def lr_scheduler_fn(
     raise ValueError(f"Unknown scheduler `{learning_rate_scheduler}`.")
 
 
-def metrics_fn() -> Dict:
-    return {"accuracy": Accuracy()}
+def metrics_fn(num_outputs: int) -> Dict:
+    task = "binary" if num_outputs == 2 else "multiclass"
```
lballes, line 363: I don't think this will work. IIUC, the binary accuracy in `torchmetrics` expects scalar predictions.

examples/getting_started/renate_config.py
```diff
@@ -93,7 +93,7 @@ def buffer_transform() -> Callable:
 
 
 def metrics_fn() -> Dict:
-    return {"accuracy": Accuracy()}
+    return {"accuracy": Accuracy(task="multiclass", num_classes=10)}
```
lballes, line 96: We need to be very careful with this new `torchmetrics` interface. It caused me some troubles in another project. By default it uses a `"macro"` average for multiclass accuracy, i.e., it computes average accuracy for each label, then averages those. We probably want the `"micro"` average, right?

lballes (changes requested):

examples/getting_started/renate_config.py
wistuba: That is correct but we are using Accuracy which uses "micro": https://torchmetrics.readthedocs.io/en/stable/classification/accuracy.html#accuracy
I've initially used the MulticlassAccuracy and then the tests failed because of the macro averaging. Now they pass, so things should work as intended.

src/renate/benchmark/experiment_config.py
wistuba: good point. do we want to use `num_outputs==1` and account for this change in `loss_fn` and `model_fn`?

examples/getting_started/renate_config.py
wistuba: ok, I've changed everything to multiclass and added the `average` argument

lballes (approved):

PR #323, opened 2023-06-30
Add Micro Average Accuracy

wistuba:

### DIFF
--- a/src/renate/benchmark/experimentation.py
+++ b/src/renate/benchmark/experimentation.py
@@ -23,6 +23,7 @@ from renate.evaluation.metrics.classification import (
     backward_transfer,
     forgetting,
     forward_transfer,
+    micro_average_accuracy,
 )
 from renate.training import run_training_job
 from renate.training.training import submit_remote_job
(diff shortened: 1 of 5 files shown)


src/renate/benchmark/experimentation.py
```diff
@@ -72,12 +75,13 @@ def cumulative_metrics_summary(
             metrics.
         cumulative_metrics: The list of (name, metric) tuples.
         num_tasks: The total number of tasks.
+        num_instances: Count of test data points for each task.
     """
     data = []
-    for task_id in range(num_tasks + 1):
+    for task_id in range(num_tasks):
```
lballes, line 81: Why did this change? I don't understand why it was `num_tasks + 1` before, but this change doesn't seem related to the micro-average accuracy.

wistuba: Agreed, not related to micro average at all, just some clean up.

In L384 the `-1` was removed so that would be the adequate change in this function.

lballes (approved):

PR #337, opened 2023-07-11
Enable Downloading Large Files

wistuba:
Current code loads downloaded file into memory first before writing to disk. This does not work for large files.

### DIFF
--- a/src/renate/utils/file.py
+++ b/src/renate/utils/file.py
@@ -305,8 +305,11 @@ def download_file(
     if src_bucket is None:
         if not os.path.exists(os.path.join(data_path, dataset_name)):
             os.makedirs(os.path.join(data_path, dataset_name))
-        response = requests.get(os.path.join(url, file_name), allow_redirects=True)
-        open(os.path.join(data_path, dataset_name, file_name), "wb").write(response.content)
+        with requests.get(os.path.join(url, file_name), allow_redirects=True, stream=True) as r:
+            r.raise_for_status()
+            with open(os.path.join(data_path, dataset_name, file_name), "wb") as f:
+                for chunk in r.iter_content(chunk_size=8192):
+                    f.write(chunk)
     else:
         download_file_from_s3(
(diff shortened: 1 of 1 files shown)


src/renate/utils/file.py
```diff
@@ -305,8 +305,11 @@ def download_file(
     if src_bucket is None:
         if not os.path.exists(os.path.join(data_path, dataset_name)):
             os.makedirs(os.path.join(data_path, dataset_name))
-        response = requests.get(os.path.join(url, file_name), allow_redirects=True)
-        open(os.path.join(data_path, dataset_name, file_name), "wb").write(response.content)
+        with requests.get(os.path.join(url, file_name), allow_redirects=True, stream=True) as r:
+            r.raise_for_status()
+            with open(os.path.join(data_path, dataset_name, file_name), "wb") as f:
+                for chunk in r.iter_content(chunk_size=8192):
```
prabhuteja12, line 311: Is the 8192 number of significance?

wesk (commented):
What happens if it fails? Is there a try/except somewhere else?

src/renate/utils/file.py
wistuba: No, we could also use 1024. Default is 1: https://requests.readthedocs.io/en/latest/api/#requests.Response.iter_content

wistuba:
> What happens if it fails? Is there a try/except somewhere else?

If it fails, it fails with one of these exceptions: https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions
What do you recommend doing in the `except` part? If we cannot download the file, we cannot continue.

prabhuteja12 (approved):

wesk:
You could set up automatic retries to handle transient failures: https://requests.readthedocs.io/en/latest/user/advanced/?highlight=retry#example-automatic-retries

But agree code is good to start with, and you can add later if it's ever a problem.

PR #338, opened 2023-07-14
Adding a Datacollator to handle the wild time text datasets

prabhuteja12:

### DIFF
--- a/src/renate/benchmark/datasets/wild_time_data.py
+++ b/src/renate/benchmark/datasets/wild_time_data.py
@@ -1,9 +1,10 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
-from transformers import PreTrainedTokenizer
+from transformers import BatchEncoding, DataCollatorWithPadding, PreTrainedTokenizer
 
 from renate import defaults
 from renate.data.data_module import RenateDataModule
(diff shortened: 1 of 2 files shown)


src/renate/utils/hf_utils.py
```diff
@@ -0,0 +1,44 @@
...
+# SPDX-License-Identifier: Apache-2.0
+from dataclasses import dataclass
+from typing import Any, Dict, List
+
+from transformers import BatchEncoding, DataCollatorWithPadding
+
+
+@dataclass
+class DataCollatorWithPaddingForWildTime(DataCollatorWithPadding):
+    # This adds to the `transformer` library's DataCollatorWithPadding. That data collator expects
```
wistuba, line 11: make this a docstring

test/renate/utils/test_hf_utils.py
```diff
@@ -0,0 +1,51 @@
...
+            True,
+        ),
+    ],
+)
+@pytest.mark.parametrize("tokenizer", [AutoTokenizer.from_pretrained("bert-base-cased")])
+def test_serialize_wildtime_collator(features, shape, error, tokenizer):
+    if not error:
+        collator = DataCollatorWithPaddingForWildTime(tokenizer)
+        out = collator(features)[0]
+        print(out)
```
wistuba, line 43: remove print

test/renate/utils/test_hf_utils.py
```diff
@@ -0,0 +1,51 @@
...
+            [
+                ({"input_ids": [1, 2, 4]}, 0, {}),
+            ],
+            torch.Size([1, 3]),
+            True,
+        ),
+    ],
+)
+@pytest.mark.parametrize("tokenizer", [AutoTokenizer.from_pretrained("bert-base-cased")])
+def test_serialize_wildtime_collator(features, shape, error, tokenizer):
```
wistuba, line 39: this doesn't seem to cover the case where metadata is not an empty dict

wistuba (commented):
did you try to run something like DER to test whether it works with a buffer?

src/renate/utils/hf_utils.py
prabhuteja12: It doesn't adhere to the traditional docstring convention of a summary + detail separation.

test/renate/utils/test_hf_utils.py
prabhuteja12: added one. the tests are quite contrived to begin with.

src/renate/utils/hf_utils.py
wistuba: Should be possible to reformat it that way, right?

test/renate/utils/test_hf_utils.py
```diff
@@ -0,0 +1,59 @@
...
+                (({"input_ids": [1, 2, 4]}, 0), {}),
+                (({"input_ids": [4, 5, 6, 7, 8, 9]}, 1), {}),
+            ],
+            torch.Size([2, 6]),
+            False,
+        ),
+        (
+            [
+                (({"input_ids": [1, 2, 4]}, 0), [1, 2]),
+                (({"input_ids": [4, 5, 6, 7, 8, 9]}, 1), [5, 6]),
```
wistuba, line 32: is `[5, 6]` the metadata? If yes, that's not the right format. It will be a dict.

wistuba (approved):

PR #339, opened 2023-07-16
Fix Scenario for CLEAR

wistuba:
- Fixes the problem we observed with CLEAR (the problem was that the scenario always return the same data chunk)
- Robust loading of the CLEAR dataset and fixing minor inaccuracies.
- Refactoring such that all time-incremental datasets (CLEAR, wild-time-data) use the same scenario


doc/benchmarking/renate_benchmarks.rst
```diff
@@ -184,11 +184,8 @@ The first part contains all instances with classes 1 and 2, the second with clas
     * - Scenario Name
       - Description
       - Settings
-    * - :py:class:`~renate.benchmark.scenarios.BenchmarkScenario`
-      - Used in combination only with CLEAR-10 or CLEAR-100.
-      - * :code:`num_tasks`: Number of data partitions.
-    * - :py:class:`~renate.benchmark.scenarios.WildTimeScenario`
-      - Used in combination only with Wild-Time datasets. This is not the scenario used in the paper.
+    * - :py:class:`~renate.benchmark.scenarios.TimeIncrementalScenario`
+      - Used in combination only with Wild-Time datasets or CLEAR. This is not the scenario used in the paper.
```
610v4nn1, line 188: nit: in which paper? both?

src/renate/benchmark/experiment_config.py
```diff
@@ -331,7 +333,9 @@ def test_transform(dataset_name: str) -> Optional[Callable]:
     if dataset_name in ["CLEAR10", "CLEAR100"]:
         return transforms.Compose(
             [
-                transforms.Resize(224),
+                transforms.Resize(
```
610v4nn1, line 336: is this following the pre-processing described in the CLEAR paper?

610v4nn1 (approved):

src/renate/benchmark/experiment_config.py
wistuba: In the paper, they work on PIL images. Then this is not required. We work on Tensors. According to the documentation, this should bring it closer to what you would do when working on PIL images. Check the part on `antialiasing`: https://pytorch.org/vision/main/generated/torchvision.transforms.Resize.html

doc/benchmarking/renate_benchmarks.rst
wistuba: updated it

610v4nn1 (approved):

PR #345, opened 2023-07-21
Fixing Bug with HPO

wistuba:
Hyperparameter defined as part of the components were overwritten by values in `state_dict`. To address this issue, `Component` no longer is a `nn.Module`.

Moving from torch.tensor to Python floats had led to changes for CLS (and therefore Super-ER) at the 8th digit. These errors accumulate and therefore give some different numbers. To account for this, I've changed the expected values for CLS and Super-ER.


src/renate/updaters/learner_components/losses.py
```diff
@@ -376,68 +276,42 @@ def _loss(
...
+        """Load relevant information from checkpoint."""
+        super().on_load_checkpoint(checkpoint)
+        self._plastic_model = checkpoint["component-cls-plastic-model"]
+        self._stable_model = checkpoint["component-cls-stable-model"]
+        self._iteration = checkpoint["component-cls-iteration"]
+
+    def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
+        """Add plastic and stable model to checkpoint."""
+        super().on_save_checkpoint(checkpoint)
+        checkpoint["component-cls-plastic-model"] = self._plastic_model
```
lballes, line 315: We should save and load `self._plastic_model.state_dict()` instead of the model object itself.

test/integration_tests/configs/suites/quick/cls-er.json
```diff
@@ -5,6 +5,6 @@
   "dataset": "mnist.json",
   "backend": "local",
   "job_name": "class-incremental-mlp-cls-er",
-  "expected_accuracy_linux": [[0.9839243292808533, 0.9740450382232666]],
+  "expected_accuracy_linux": [[0.9834515452384949, 0.9740450382232666]],
```
lballes, line 8: Shouldn't the results remain unaffected by the changes?

lballes (changes requested):

test/integration_tests/configs/suites/quick/cls-er.json
wistuba: See PR description for more details. In short, working with torch.tensor(1.0) and float(1.0) seem to cause differences.

src/renate/updaters/learner_components/losses.py
wistuba: done

PR #356, opened 2023-07-31
Experimentation Tools

wistuba:
Experimentation tools to run experiment. Adds configurations to run experiments with CLEAR10/100 and DomainNet.

It also upgrades the SageMaker defaults to PyTorch 1.13.1 and Python 3.9.


lballes (approved):
Thanks for setting this up. This will be hugely helpful!

PR #358, opened 2023-08-01
Cleanup iCarl

wistuba:
iCarl is no replay-based method and therefore doesn't have to be a ReplayLearner nor expect its arguments.


610v4nn1:
iCarl is replay-based but for our hierarchy of classes it falls under the Avalanche-based learners and does not need to inherit from replay learners

PR #359, opened 2023-08-02
Replace memory batch size with a fraction of the total batch size

wistuba:
New attempt of #217 


lballes (approved):
Not sure why it works now, but LGTM.

In this version, offline ER still uses only `batch_size` in the initial update rather than `batch_size + memory_batch_size`, but I can adapt this in a separate PR.

PR #362, opened 2023-08-04
Support Use of Joint and GDumb with Pre-Trained Models

wistuba:
Joint and Gdumb reset the models `on_model_update_start`. This results in a non-desired behavior when working with pre-trained models.
There is also no option to *not* reset the model at all.

The change introduces a new flag `reset` which allows to control whether the model will be reset. Furthermore, in case of a reset, the pre-trained model will be reloaded instead of using an untrained model.

The integration tests are expected to break for Joint and GDumb due to a different initialization of the model. Before, the workflow was creating the model and resetting it. Now, it only creates the model without reset.


prabhuteja12 (commented):
Documenting an alternative to this approach: 

The resetting is method + network specific. Thus an alternative is to define a `reset_parameters` for `RenateBenchmarkingModule` and overload them only for specific modules (ViT, Text transformers etc). The current `on_model_training_start` would invoke the this `reset_parameters` to reset it.

prabhuteja12 (approved):

PR #363, opened 2023-08-07
Wild Time Benchmarks and Small Memory Hack

wistuba:
Add files required to run experimentation with Wild Time Benchmarks.

Save PIL images rather than tensors in buffers to keep memory space lower.

Found bug: `optimizer_fn` doesn't work as intended. If it exists, it must return an optimizer. We would like that it can exist but it either returns an optimizer or `None`. Currently "fixed" by adding all arguments expected for an optimizer to the function.

### DIFF
--- a/benchmarks/experiment_configs/datasets/arxiv.json
+++ b/benchmarks/experiment_configs/datasets/arxiv.json
@@ -1,0 +1,6 @@
+{
+  "dataset_name": "arxiv",
+  "src_bucket": "my_bucket",
+  "src_object_name": "dataset/wildtime/arxiv.hdf5",
+  "num_outputs": 172
+}
--- a/benchmarks/experiment_configs/scenarios/arxiv-16updates.json
+++ b/benchmarks/experiment_configs/scenarios/arxiv-16updates.json
@@ -1,0 +1,9 @@
+{
+  "val_size": 0.1,
+  "scenario_name": "DataIncrementalScenario",
+  "num_tasks": 16,
+  "max_epochs": 2,
+  "optimizer": "AdamW",
+  "learning_rate": 0.00002,
+  "weight_decay": 0.01
+}
--- a/src/renate/benchmark/datasets/wild_time_data.py
+++ b/src/renate/benchmark/datasets/wild_time_data.py
@@ -90,6 +90,7 @@ class WildTimeDataModule(DataIncrementalDataModule):
             "time_step": available_time_steps(self._dataset_name)[self.data_id],
             "data_dir": self._data_path,
             "in_memory": self._dataset_name != "fmow",
+            "transform": None if self._dataset_name != "fmow" else lambda x: x,
         }
         if self._tokenizer:
             kwargs["transform"] = lambda x: self._tokenizer(x, **(self._tokenizer_kwargs or {}))
--- a/src/renate/benchmark/experiment_config.py
+++ b/src/renate/benchmark/experiment_config.py
@@ -5,11 +5,12 @@ from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import torch
 import wild_time_data
-from torch.optim import Optimizer
+from torch.optim import AdamW, Optimizer
 from torch.optim.lr_scheduler import CosineAnnealingLR, StepLR, _LRScheduler
 from torchmetrics.classification import MulticlassAccuracy
 from torchvision.transforms import transforms
 from transformers import AutoTokenizer
+from wild_time_data.datasets import FMoW
 
 from renate.benchmark.datasets.nlp_datasets import HuggingFaceTextDataModule
 from renate.benchmark.datasets.vision_datasets import (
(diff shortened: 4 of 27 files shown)


benchmarks/experiment_configs/datasets/arxiv.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dataset_name": "arxiv",
+  "src_bucket": "my_bucket",
```
prabhuteja12, line 3: Is this accessible to the team/public?

src/renate/benchmark/datasets/wild_time_data.py
```diff
@@ -90,6 +90,7 @@ def setup(self) -> None:
             "time_step": available_time_steps(self._dataset_name)[self.data_id],
             "data_dir": self._data_path,
             "in_memory": self._dataset_name != "fmow",
+            "transform": None if self._dataset_name != "fmow" else lambda x: x,
```
prabhuteja12, line 93: Can you try this on multigpu? I reckon this DDP won't like this lambda.

src/renate/benchmark/experiment_config.py
```diff
@@ -317,6 +318,8 @@ def _get_normalize_transform(dataset_name):
 
 def train_transform(dataset_name: str) -> Optional[Callable]:
     """Returns a transform function to be used in the training."""
+    if dataset_name == "fmow":
+        return FMoW.default_transform
```
prabhuteja12, line 322: There seems to exist a `default_transform(datasetname)` in the package. Any reason to go this route?

src/renate/benchmark/experiment_config.py
```diff
@@ -398,3 +405,13 @@ def lr_scheduler_fn(
...
+
+
+def optimizer_fn(
+    optimizer: str,
+    learning_rate: float,
+    weight_decay: float,
+    momentum: float = 0.0,  # TODO: fix problem that occurs when removing this
+) -> Callable:
+    if optimizer == "AdamW":
+        return partial(AdamW, lr=learning_rate, weight_decay=weight_decay)
```
prabhuteja12, line 417: can this written as: 
`partial(getattr(torch.optim, optimizer),  lr=learning_rate, weight_decay=weight_decay)`? Why the specific handling of AdamW?

test/renate/benchmark/test_experimentation_config.py
```diff
@@ -363,3 +365,17 @@ def test_loss_fn_returns_correct_reduction_type():
...
+    partial_optimizer = optimizer_fn(
+        optimizer="AdamW", learning_rate=expected_learning_rate, weight_decay=0.1
+    )
+    optimizer: AdamW = partial_optimizer(Linear(10, 10).parameters())
+    assert isinstance(optimizer, AdamW)
+    assert optimizer.defaults["lr"] == expected_learning_rate
+
+
+def test_optimizer_fn_unknown_optimizer():
+    assert optimizer_fn(optimizer="UNKNOWN_OPTIMIZER", learning_rate=0.1, weight_decay=1) is None
```
prabhuteja12, line 381: What does this test aim at accomplishing? Would throwing an error be better?

benchmarks/experiment_configs/scenarios/arxiv-16updates.json
```diff
@@ -0,0 +1,9 @@
+{
+  "val_size": 0.1,
+  "scenario_name": "DataIncrementalScenario",
+  "num_tasks": 16,
+  "max_epochs": 2,
+  "optimizer": "AdamW",
+  "learning_rate": 0.00002,
```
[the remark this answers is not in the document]
wistuba: wild time paper

src/renate/benchmark/datasets/wild_time_data.py
wistuba: you mean, the function needs to be pickable?

prabhuteja12: yes

src/renate/benchmark/experiment_config.py
wistuba: in principle there is no problem if we make it more general. however, how would that work with SGD + momentum? right now, this logic will only be triggered for AdamW and otherwise we will fall back to the standard optimizers (SGD, Adam), we have in the library.

test/renate/benchmark/test_experimentation_config.py
wistuba: similar to above. returning `None` is fine. This will be the case when using SGD or Adam. Raising an error in `optimizer_fn` should not happen. If the input is wrong, it will eventually be triggered at a different part of the code.

src/renate/benchmark/experiment_config.py
wistuba: good point

src/renate/benchmark/datasets/wild_time_data.py
wistuba: I've just noticed that 3 lines below is yet another lambda function...

wistuba: I've tested it, it doesn't cause any problems.

prabhuteja12 (approved):

PR #364, opened 2023-08-08
Masking of logits of irrelevant classes

prabhuteja12:
In the specific case of class incremental learning, the current implication trains the weights of the classes (even the ones that are not part of the current task). This PR creates a new flag that fills the "to be ignored" logits with -inf. 

This method is currently not implemented for Avalanche learners. 


src/renate/updaters/learner.py
```diff
@@ -76,6 +83,18 @@ def __init__(
         self._rng = get_generator(self._seed)
         self.save_hyperparameters(ignore=self._ignored_hyperparameters())
 
+    def _possibly_populate_class_mask(self):
+        # Compute which logits to mask, if the model has a _num_outputs attribute. This is possible
+        # when self._model is a RenateBenchmarkingModule. It its not, the first forward prop will
+        # populate the _class_mask.
+        self._classes_in_current_task = unique_classes(self._train_dataset)
+        warnings.warn(f"{self._classes_in_current_task}")
```
wistuba, line 91: is this a debugging left-over? anyway, this should not be a `warning.warn` call. Instead, it should be output generated by `logging`. It probably also does not deserve warning level. If you want to leave it, please provide a bit more context in the logging message.

src/renate/utils/pytorch.py
```diff
@@ -125,3 +125,28 @@ def cat_nested_tensors(
...
+
+
+def complementary_indices(num_outputs: int, valid_classes: Set[int]) -> List[int]:
+    """Compute the asymmetric difference between the two arguments
+
+    Args:
+        num_outputs: An integer of total number of classes the model can output.
+        valid_classes: A set of integers of valid classes.
+    """
+    return list(set(range(num_outputs)) - valid_classes)
```
wistuba, line 152: nit: `[class_idx for class_idx in range(num_outputs) if class_idx not in valid_classes]`

src/renate/updaters/learner.py
```diff
@@ -76,6 +83,18 @@ def __init__(
...
+        if hasattr(self._model, "_num_outputs"):
```
wistuba, line 92: any advantage of adding this special edge case for `RenateBenchmarkingModule`?

wistuba (commented):
as far as i can see, when no class masks are used, the code does not change. that's good because we do not want to assume a classification task again.

src/renate/updaters/learner.py
prabhuteja12: For debugging purposes. Now removed.

src/renate/utils/pytorch.py
prabhuteja12: ok

src/renate/training/training.py
```diff
@@ -623,7 +627,7 @@ def _execute_training_and_tuning_job_locally(
             + "\n\nLogs (stdout):\n\n{}".format("".join(backend.stdout(0)))
             + "\n\nLogs (stderr):\n\n{}".format("".join(backend.stderr(0)))
         )
-
+    print("".join(backend.stdout(0)))
```
wistuba, line 630: left over from debugging

prabhuteja12: cleaned.

src/renate/updaters/learner.py
```diff
@@ -192,6 +207,16 @@ def training_step(
...
+        if self._mask_unused_classes:
+            if self._class_mask is None:
+                # Now is the time to repopulate the class_mask
+                self._class_mask = torch.tensor(
+                    complementary_indices(outputs.size(1), self._classes_in_current_task),
+                    device=outputs.device,
+                    dtype=torch.long,
+                )
+            # fill the logits with -inf
+            outputs.index_fill_(1, self._class_mask.to(outputs.device), -float("inf"))
```
wistuba, line 219: can we make all of the new code a function and only call this function here? then we can also reuse it for Offline-ER (new data and buffer data)

src/renate/updaters/avalanche/model_updater.py
```diff
@@ -276,6 +276,7 @@ def __init__(
         deterministic_trainer: bool = defaults.DETERMINISTIC_TRAINER,
         gradient_clip_val: Optional[float] = defaults.GRADIENT_CLIP_VAL,
         gradient_clip_algorithm: Optional[str] = defaults.GRADIENT_CLIP_ALGORITHM,
+        mask_unused_classes: bool = defaults.MASK_UNUSED_CLASSES,
```
wistuba, line 279: maybe a warning somewhere that this is currently not supported?

src/renate/updaters/learner.py
```diff
@@ -76,6 +84,11 @@ def __init__(
         self._rng = get_generator(self._seed)
         self.save_hyperparameters(ignore=self._ignored_hyperparameters())
 
+    def _find_unique_classes_in_task(self):
```
wistuba, line 87: personally, I prefer if functions don't have side effects. Given that nothing is really happening in this function, maybe we can drop it and use `unique_classes` directly?

wistuba (changes requested):
I see some code for computing the mask in Offline-ER. I don't see anything similar to that in the ReplayLearner. Is this change already working with ER, Joint, and GDumb?

src/renate/updaters/avalanche/model_updater.py
prabhuteja12: avalanche?

src/renate/updaters/learner.py
prabhuteja12: Done.

src/renate/updaters/avalanche/model_updater.py
prabhuteja12: Logged a warning. Is a `warnings.warn` preferred?

wistuba: they have different semantic meanings. warnings.warn is for warnings that can be fixed by the user, i.e., by changing the input. the warning from logging is for warnings that the user is not responsible for.

src/renate/updaters/avalanche/model_updater.py
```diff
@@ -47,6 +47,13 @@
 class AvalancheModelUpdater(SingleTrainingLoopUpdater):
     _report = Reporter()
 
+    def __init__(self, *args, **kwargs):
+        if kwargs.get("mask_unused_classes", False) is True:
+            logger.warning(
```
wistuba, line 52: `warnings.warn`

src/renate/updaters/learner.py
```diff
@@ -44,6 +45,8 @@ class RenateLightningModule(LightningModule, abc.ABC):
         batch_size: Training batch size.
         logged_metrics: Metrics logged additional to the default ones.
         seed: See :func:`renate.models.utils.get_generator`.
+        mask_unused_classes: Flag to use if logits corresponding to unused classes are to be killed.
```
wistuba, line 48: killed -> ignored

src/renate/utils/misc.py
```diff
@@ -12,3 +16,24 @@ def int_or_str(x: str) -> Union[str, int]:
         return int(x)
     except ValueError:
         return x
+
+
+def possibly_populate_mask_and_kill_logits(
```
wistuba, line 21: can we rename the function? I would refrain from "martial language". does replacing "kill" with "ignore work for you?.

at different places we add the prefix "maybe" to indicate that the function is potentially executed. Maybe we can follow this convention? (possibly -> maybe)

wistuba (changes requested):

src/renate/utils/misc.py
prabhuteja12: Changed.

src/renate/utils/misc.py
```diff
@@ -12,3 +16,24 @@ def int_or_str(x: str) -> Union[str, int]:
...
         return x
+
+
+def maybe_populate_mask_and_ignore_logits(
+    use_masking: bool,
+    class_mask: Optional[torch.Tensor],
+    classes_in_current_task: Optional[Set[int]],
+    logits: torch.Tensor,
+):
+    """Snippet to compute which logits to kill after computing the class mask if required."""
```
wistuba, line 27: "to ignore after"

wistuba (commented):
I'd like to update one more docstring, then we can merge

wistuba (approved):

PR #365, opened 2023-08-08
Clean Up Learner Checkpoint and Fix Model Loading

wistuba:
Parameters stored in the Learner checkpoint were overwriting parameters of the passed model.

Numbers for Joint/GDumb are expected to change since before these methods were initialized with the weights of the previous model.

### DIFF
--- a/src/renate/updaters/model_updater.py
+++ b/src/renate/updaters/model_updater.py
@@ -142,12 +142,12 @@ class RenateModelCheckpoint(ModelCheckpoint):
 
         # Finalize model update.
         pl_module.on_model_update_end()
+        # Overwrite checkpoint.
+        self._save_checkpoint(trainer, str(learner_state_path))
         # Save permanently.
         if trainer.is_global_zero:
-            # Save the buffer only on rank zero.
+            # Save Learner only on rank zero.
             pl_module.save(self._output_state_folder)
-        # Overwrite checkpoint.
-        self._save_checkpoint(trainer, learner_state_path)
 
     def teardown(self, trainer: Trainer, pl_module: LightningModule, stage: str) -> None:
         """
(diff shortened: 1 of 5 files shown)


src/renate/updaters/model_updater.py
```diff
@@ -187,10 +186,14 @@ def teardown(self, trainer: Trainer, pl_module: LightningModule, stage: str) ->
...
-                state_dict = torch.load(learner_state_path)["state_dict"]
-                out_sd = {k.replace("_model.", "", 1): v for k, v in state_dict.items()}
-                # Replace only 1 instance because we have to load it into RenateModule.
+                learner_state = torch.load(learner_state_path)
+                out_sd = {
+                    k.replace("_model.", "", 1): v for k, v in learner_state["state_dict"].items()
+                }  # Replace only 1 instance because we have to load it into RenateModule.
                 torch.save(out_sd, defaults.model_file(self.dirpath))
+                # Remove model from learner checkpoint
+                learner_state["state_dict"] = {}
```
prabhuteja12, line 195: Why empty? `learner_state.pop("state_dict")` instead?

src/renate/updaters/model_updater.py
```diff
@@ -382,6 +385,7 @@ def _load_learner(
             self._learner_state_file,
             model=self._model,
             logged_metrics=self._logged_metrics,
+            strict=False,
```
prabhuteja12, line 388: Why?

prabhuteja12 (changes requested):

src/renate/updaters/model_updater.py
wistuba: If it is empty, loading with `strict=False` will make basically no changes. The key must exist since lightning explicitly accesses it.

src/renate/updaters/model_updater.py
wistuba: Let me check if it is necessary. But we still try to load the checkpoint and since keys are missing in `state_dict`, it complains otherwise.

prabhuteja12 (approved):

PR #368, opened 2023-08-10
Enable Custom Grouping for DataIncrementalScenario

wistuba:
Rename `class_groupings` to `groupings` in `ClassIncrementalScenario`.

`DataIncrementalScenario` has now an argument `groupings` that allows grouping datasets similar to the `groupings` in `ClassIncrementalScenario` for classes.

Added support for training on `yearbook` with ViT.

Functionality exposed for benchmarking.

### DIFF
--- a/benchmarks/run_benchmark.py
+++ b/benchmarks/run_benchmark.py
@@ -112,6 +112,7 @@ if __name__ == "__main__":
             instance_type="ml.g4dn.xlarge",
             n_workers=1,
             max_time=args.max_time,
+            instance_max_time=args.max_time,
             seed=seed,
             job_name=args.job_name[:36],
             devices=1,
--- a/src/renate/benchmark/experiment_config.py
+++ b/src/renate/benchmark/experiment_config.py
@@ -11,7 +11,6 @@ from torchmetrics.classification import MulticlassAccuracy
 from torchvision.transforms import transforms
 from transformers import AutoTokenizer
 from wild_time_data import default_transform
-from wild_time_data.datasets import FMoW
 
 from renate.benchmark.datasets.nlp_datasets import HuggingFaceTextDataModule
 from renate.benchmark.datasets.vision_datasets import (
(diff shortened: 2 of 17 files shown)


src/renate/benchmark/experiment_config.py
```diff
@@ -337,21 +355,38 @@ def train_transform(dataset_name: str) -> Optional[Callable]:
     if dataset_name in ["CLEAR10", "CLEAR100", "DomainNet"]:
         return transforms.Compose(
             [
-                transforms.ToTensor(),
-                transforms.Resize(
-                    224, interpolation=transforms.InterpolationMode.BILINEAR, antialias=True
-                ),
+                transforms.Resize(224),
```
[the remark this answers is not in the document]
wistuba: the input to resize is no longer a tensor but a pilimage. so this should be default: https://pytorch.org/vision/main/generated/torchvision.transforms.Resize.html

benchmarks/run_benchmark.py
```diff
@@ -112,6 +112,7 @@ def load_config(scenario_file, model_file, updater_file, dataset_file):
             instance_type="ml.g4dn.xlarge",
             n_workers=1,
             max_time=args.max_time,
+            instance_max_time=args.max_time,
```
prabhuteja12, line 115: difference to `max_time`?

wistuba: max_time refers to the tuning time per update step. instance_max_time refers to after which time the sagemaker training job is aborted.

test/renate/benchmark/test_scenarios.py
```diff
@@ -49,42 +49,107 @@ def test_failing_to_init(tmpdir, scenario_cls, kwargs):
...
+def test_data_incremental_scenario_grouping():
+    """Test whether grouping in DataIncrementalScenario works"""
+    scenario = DataIncrementalScenario(
+        data_module=DummyDataIncrementalDataModule(0, (10, 1), val_size=0.2),
+        chunk_id=0,
+        groupings=((0, 1), (2,)),
+    )
+    scenario.prepare_data()
+    scenario.setup()
+    assert set(int(scenario.train_data()[i][1]) for i in range(80)) == {0}
```
prabhuteja12, line 91: Can you write a comment about what is being tested for, here? I don't follow the 80, and following 20 are.

prabhuteja12 (approved):

PR #379, opened 2023-08-16
Modifies current text transformer implementation to a RenateBenchmarkingModule

prabhuteja12:


requirements.txt
```diff
@@ -13,5 +13,5 @@ torchmetrics>=0.11.0, <0.11.5
 torchvision>=0.13.0, <0.15.2
 deepspeed==0.9.1
 datasets>=2.9.0, <2.14.1
-transformers>=4.30.0, <4.30.2
+transformers>=4.31.0
```
wistuba, line 16: please add `, <4.31.1`

src/renate/benchmark/models/transformer.py
```diff
@@ -4,36 +4,56 @@
...
-    def _add_task_params(self, task_id: str) -> None:
-        assert not len(self._tasks_params_ids), "Transformer does not work for multiple tasks."
+        self._backbone = model
+
+    def forward(self, x: Dict[str, Tensor], task_id: str = defaults.TASK_ID) -> torch.Tensor:
+        out = self.get_backbone(task_id=task_id)(**x, return_dict=True)
+        if hasattr(out, "pooler_output"):
+            x = out.pooler_output
+        else:
+            x = out.last_hidden_state[:, 0]  # 0th element is used for classification.
```
wistuba, line 52: not sure about this, want to hear your opinion: what do you think of adding this logic somehow to the object returned by `get_backbone` such that overriding forward is not necessary?

test/renate/benchmark/models/test_text_transformer.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@pytest.mark.parametrize(
+    "model_name,input_dim",
+    [
+        ["distilbert-base-uncased", (128,)],
+        ["bert-base-uncased", (256,)],
+    ],
+)
+def test_text_transformer_fwd(model_name, input_dim):
```
wistuba, line 19: if we do not change the `get_backbone` above, can we also test with `prediction_strategy` active?

requirements.txt
prabhuteja12: Done.

wistuba (approved):

PR #381, opened 2023-08-17
Make offline ER us total batch size in first update

lballes:
Offline ER currently uses only `batch_size` in the first model update. This PR makes it use `batch_size + memory_batch_size` to be comparable to joint training, et cetera. 

### DIFF
--- a/src/renate/updaters/experimental/offline_er.py
+++ b/src/renate/updaters/experimental/offline_er.py
@@ -85,20 +85,10 @@ class OfflineExperienceReplayLearner(ReplayLearner):
                 collate_fn=self._train_collate_fn,
             )
         else:
-            # Manually create a dataloader for the current task with total batch size.
-            train_dataset = _TransformedDataset(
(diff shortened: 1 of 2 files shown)


src/renate/updaters/experimental/offline_er.py
```diff
@@ -83,6 +84,21 @@ def train_dataloader(self) -> DataLoader:
                 pin_memory=True,
                 collate_fn=self._train_collate_fn,
             )
+        else:
+            # Manually create a dataloader for the current task with total batch size.
```
wistuba, line 88: Can we reuse the code instead of overriding it?

```
self._batch_size += self._memory_batch_size
loaders["current_task"] = super().train_dataloader()
```

lballes: I can do something like that if you prefer it. We would have to undo the change after creating the data loader though.

wistuba (approved):

PR #390, opened 2023-08-23
Refactor Offline-ER to work with `collate_fn`

wistuba:
Offline-ER applies collate_fn individually on new and memory data. This change will apply the collate function on the entire batch instead.

### DIFF
--- a/src/renate/utils/pytorch.py
+++ b/src/renate/utils/pytorch.py
@@ -195,23 +195,26 @@ class ConcatRandomSampler(BatchSampler):
         )
 
     def __iter__(self) -> Iterator[List[int]]:
+        """Creates a batch with groups of indices where each group corresponds to one dataset."""
         if self.complete_dataset_iteration is None:
+            # Default case is iterating once over the shortest iterator. Works nicely with zip.
             for samples in zip(*self.subset_samplers):
                 yield [j for i in samples for j in i]
         else:
+            # Iterating over a specific iterator requires dealing with the length of other iterators.
             iterators = [iter(sampler) for sampler in self.subset_samplers]
             for s in iterators[self.complete_dataset_iteration]:
                 samples = []
                 for i, iterator in enumerate(iterators):
                     if i != self.complete_dataset_iteration:
                         try:
-                            samples.append(next(iterator))
+                            samples += next(iterator)
                         except StopIteration:
                             iterators[i] = iter(self.subset_samplers[i])
-                            samples.append(next(iterators[i]))
+                            samples += next(iterators[i])
                     else:
-                        samples.append(s)
-                yield [j for i in samples for j in i]
+                        samples += s
+                yield samples
 
     def __len__(self):
         return self.length
--- a/src/renate/utils/pytorch.py
+++ b/src/renate/utils/pytorch.py
@@ -20,11 +20,11 @@ def reinitialize_model_parameters(model: torch.nn.Module) -> None:
     implementations of exotic layers. A warning is logged for modules that do not implement
     `reset_parameters()`.
 
-    The actual logic of reintializing parameters depends on the type of layer. It may affect the
+    The actual logic of reinitializing parameters depends on the type of layer. It may affect the
     module's buffers (non-trainable parameters, e.g., batch norm stats) as well.
 
     Args:
-        model: The model to be re-initialized.
+        model: The model to be reinitialized.
     """
     for module in model.modules():
         # Skip modules without any parameters of their own.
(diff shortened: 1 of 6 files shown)


src/renate/utils/pytorch.py
```diff
@@ -156,3 +156,62 @@ def complementary_indices(num_outputs: int, valid_classes: Set[int]) -> List[int
...
+            else num_batches[self.complete_dataset_iteration]
+        )
+
+    def __iter__(self) -> Iterator[List[int]]:
+        if self.complete_dataset_iteration is None:
+            for samples in zip(*self.subset_samplers):
+                yield [j for i in samples for j in i]
+        else:
+            iterators = [iter(sampler) for sampler in self.subset_samplers]
+            for s in iterators[self.complete_dataset_iteration]:
```
prabhuteja12, line 203: Is this optimized? Nested for-loops for each batch seems like a lot.

prabhuteja12 (changes requested):
Can you check this works with distributed training? That uses something like a `DistributedSampler` which also modifies the data to sample from.

src/renate/utils/pytorch.py
```diff
@@ -156,3 +156,62 @@ def complementary_indices(num_outputs: int, valid_classes: Set[int]) -> List[int
...
     return [class_idx for class_idx in range(num_outputs) if class_idx not in valid_classes]
+
+
+class ConcatRandomSampler(BatchSampler):
+    """Sampler for sampling batches from ConcatDatasets.
+
+    Args:
+        dataset_lengths: The length for the different datasets.
+        batch_sizes: Batch sizes used for specific datasets.
+        complete_dataset_iteration: Provide an index to indicate over which dataset to fully
```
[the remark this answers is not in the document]
wistuba: suggestions?

src/renate/utils/pytorch.py
wistuba: there is no nested loop for each batch. it is a single loop over each iterator. in case 1 this is hidden within `zip` but it also has a loop over each iterator and calls next.

src/renate/utils/pytorch.py
```diff
@@ -156,3 +156,76 @@ def complementary_indices(num_outputs: int, valid_classes: Set[int]) -> List[int
...
+                SubsetRandomSampler(list(range(start_idx, end_idx)), generator),
+                batch_size,
+                True,
+            )
+            self.subset_samplers.append(subset_sampler)
+            num_batches.append((end_idx - start_idx) // batch_size)
+            data_start_idx = data_end_idx
+        self.length = (
+            min(num_batches)
+            if complete_dataset_iteration is None
```
prabhuteja12, line 203: Why not `self.complete_dataset_iteration` here?

src/renate/utils/pytorch.py
```diff
@@ -156,3 +156,76 @@ def complementary_indices(num_outputs: int, valid_classes: Set[int]) -> List[int
         valid_classes: A set of integers of valid classes.
     """
     return [class_idx for class_idx in range(num_outputs) if class_idx not in valid_classes]
+
+
+class ConcatRandomSampler(BatchSampler):
```
prabhuteja12, line 161: Why inherit from BatchSampler?

src/renate/utils/pytorch.py
```diff
@@ -156,3 +156,76 @@ def complementary_indices(num_outputs: int, valid_classes: Set[int]) -> List[int
...
+        data_start_idx = 0
+        num_batches = []
+        rank = 0 if sampler is None else sampler.rank
+        num_replicas = 1 if sampler is None else sampler.num_replicas
+        for dataset_length, batch_size in zip(dataset_lengths, batch_sizes):
+            data_end_idx = data_start_idx + dataset_length
+            start_idx = data_start_idx + round(dataset_length / num_replicas * rank)
+            end_idx = data_start_idx + round(dataset_length / num_replicas * (rank + 1))
+            subset_sampler = BatchSampler(
+                SubsetRandomSampler(list(range(start_idx, end_idx)), generator),
```
prabhuteja12, line 194: Why `BatchSampler` of `SubsetRandomSampler`?

src/renate/utils/pytorch.py
```diff
@@ -156,3 +156,76 @@ def complementary_indices(num_outputs: int, valid_classes: Set[int]) -> List[int
...
+            # iterators.
+            iterators = [iter(sampler) for sampler in self.subset_samplers]
+            for s in iterators[self.complete_dataset_iteration]:
+                samples = []
+                for i, iterator in enumerate(iterators):
+                    if i != self.complete_dataset_iteration:
+                        try:
+                            samples += next(iterator)
+                        except StopIteration:
+                            iterators[i] = iter(self.subset_samplers[i])
```
prabhuteja12, line 224: Can you add comments for this? 

Also, why not https://docs.python.org/3/library/itertools.html#itertools.zip_longest ?

test/renate/utils/test_pytorch.py
```diff
@@ -11,6 +11,7 @@
 from renate.memory.buffer import ReservoirBuffer
 from renate.utils import pytorch
 from renate.utils.pytorch import (
```
prabhuteja12, line 13: Is it possible to add a `DistributedSampler` to a test?

src/renate/utils/pytorch.py
wistuba: zip_longest requires a `fill_value`. however, we don't want to use a fixed `fill_value` but rather iterate over the other data sources.

test/renate/utils/test_pytorch.py
```diff
@@ -150,3 +151,22 @@ def test_unique_classes(tmpdir, test_dataset):
         buffer.update(ds, metadata)
         predicted_unique = unique_classes(buffer)
         assert predicted_unique == set(list(range(10)))
+
+
+@pytest.mark.parametrize(
+    "complete_dataset_iteration,expected_batches", [[None, 2], [0, 7], [1, 5], [2, 2]]
```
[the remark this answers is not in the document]
wistuba: yes. it is identical to the [2, 2] case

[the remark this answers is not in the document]
prabhuteja12: So a drop_last is implicit?

src/renate/utils/pytorch.py
wistuba: BatchSampler creates batches, SubsetRandomSampler creates random ints from the provided list (List[int] vs int)

test/renate/utils/test_pytorch.py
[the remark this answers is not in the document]
wistuba: yes. improved doc

src/renate/utils/pytorch.py
wistuba: done

src/renate/utils/pytorch.py
wistuba: changed to Sampler

test/renate/utils/test_pytorch.py
wistuba: I've added a unit test for the distributed case instead

prabhuteja12 (approved):

PR #391, opened 2023-08-24
Add schedule for alpha in offline ER

lballes:
This adds the option to put alpha (the weight for the new data) on a schedule, linearly interpolating between an initial and a final value over a given number of iterations.

### DIFF
--- a/src/renate/updaters/experimental/offline_er.py
+++ b/src/renate/updaters/experimental/offline_er.py
@@ -29,21 +29,30 @@ class OfflineExperienceReplayLearner(ReplayLearner):
     terminated.
 
     Args:
-        loss_weight_new_data: The training loss will be a convex combination of the loss on the new
-            data and the loss on the memory data. If a float (needs to be in [0, 1]) is given here,
-            it will be used as the weight for the new data. If `None`, the weight will be set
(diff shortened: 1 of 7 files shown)


src/renate/updaters/experimental/offline_er.py
```diff
@@ -29,21 +29,30 @@ class OfflineExperienceReplayLearner(ReplayLearner):
...
-            `i` and the current task is `t`.
+        alpha: The training loss will be a convex combination of the loss on the new data and the
+            loss on the memory data. If a float (needs to be in [0, 1]) is given here, it will be
+            used as the weight for the new data. If `None`, the weight will be set dynamically to
+            `N_t / sum([N_1, ..., N_t])`, where `N_i` denotes the size of task/chunk `i` and the
+            current task is `t`.
+        alpha_schedule_init: If this is given, alpha (the weight for the new data) will be on a
+            schedule starting at `alpha_schedule_init` and linearly going towards `alpha` over
+            `alpha_schedule_steps` iterations.
+        alpha_schedule_steps: Number of steps over which `alpha` is adjusted.
```
wistuba, line 40: is there a way to define this as fraction of entire training budget?

lballes: As far as I can tell, te `Learner` does not have access to the `max_epochs`.

PR #420, opened 2023-09-15
Abstracting prompting transformer for use in L2P and S-Prompt

prabhuteja12:
The previous version of L2P had the "prompting" mechanism tightly coupled. This PR separates the Prompting and the prompt selection strategy. 

### DIFF
--- a/src/renate/benchmark/models/l2p.py
+++ b/src/renate/benchmark/models/l2p.py
@@ -108,6 +108,123 @@ class PromptPool(nn.Module):
         return selected_prompts, loss_value
 
 
+class PromptedTransformer(nn.Module):
+    """This generic module is the basic prompted transformer. It takes in a model string and creates
+    the appropriate transformer. If not prompted, it returns features, and if prompted, it returns
+    the full feature using those prompts and the input image/text.
+
+    Args:
+        pretrained_model_name_or_path: A string that denotes which pretrained model from the HF hub
+            to use.
+        num_outputs: Size of the output.
+        prediction_strategy: Continual learning strategies may alter the prediction at train or test
+            time.
+        add_icarl_class_means: If ``True``, additional parameters used only by the
+            ``ICaRLModelUpdater`` are added. Only required when using that updater.
+    """
+
+    def __init__(
+        self,
+        pretrained_model_name_or_path="google/vit-base-patch16-224",
+        image_size: int = 32,
+        patch_size: int = 4,
+        num_layers: int = 12,
+        num_heads: int = 12,
+        hidden_dim: int = 768,
+        mlp_dim: int = 3072,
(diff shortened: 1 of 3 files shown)


src/renate/benchmark/models/l2p.py
```diff
@@ -108,6 +108,123 @@ def forward(self, x: torch.Tensor, manual_prompt_indices: Optional[torch.LongTen
...
+        """
+        if prompt is None:
+            return (
+                self.transformer.get_features(x)
+                if self._is_text_transformer
+                else self.transformer.get_features(x, cls_feat=cls_feat)
+            )
+            # text transformers dont support cls_feat.
+        else:
+            if self._is_text_transformer:
```
wistuba, line 197: we can flatten the if else(if else) part to if, elif, else

src/renate/benchmark/models/l2p.py
```diff
@@ -166,34 +283,22 @@ def __init__(
...
+            mlp_dim=mlp_dim,
+            dropout=dropout,
+            attention_dropout=attention_dropout,
+            num_outputs=num_outputs,
+            add_icarl_class_means=add_icarl_class_means,
+            prediction_strategy=prediction_strategy,
+        )
         prompter = PromptPool(
-            embedding_dim=transformer._embedding_size,
+            embedding_dim=transformer.transformer._embedding_size,
```
wistuba, line 301: in the following, we access a lot of protected attributes of the transformer. do we want to keep it that way or rather make them public?

src/renate/benchmark/models/l2p.py
prabhuteja12: Fixed in d8ef1c81da1e7e1e714a7ef96ab8beb775f0bad4

src/renate/benchmark/models/l2p.py
```diff
@@ -108,6 +108,123 @@ def forward(self, x: torch.Tensor, manual_prompt_indices: Optional[torch.LongTen
         return selected_prompts, loss_value
 
 
+class PromptedTransformer(nn.Module):
+    """This generic module is the basic prompted transformer. It takes in a model string and creates
+    the appropriate transformer. If not prompted, it returns features, and if prompted, it returns
+    the full feature using those prompts and the input image/text.
```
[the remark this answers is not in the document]
prabhuteja12: Fixed in 3896135f37f3e71f424265c36d75025819986c8b.

src/renate/benchmark/models/l2p.py
prabhuteja12: Fixed in 3896135f37f3e71f424265c36d75025819986c8b

wistuba (approved):

PR #441, opened 2023-09-28
CLEAR and TinyImageNet tests bugfixes

prabhuteja12:
A few tests for datasets fail due to modified interfaces etc. This PR fixes those.

### DIFF
--- a/test/renate/data/test_data_module.py
+++ b/test/renate/data/test_data_module.py
@@ -91,8 +91,8 @@ def test_torchvision_data_module(tmpdir, dataset_name, num_tr, num_te, x_shape):
 @pytest.mark.parametrize(
     "dataset_name,chunk_id,num_tr,num_te",
     [
-        ("CLEAR10", 0, 2986, 500),
-        ("CLEAR100", 0, 9945, 4984),
+        ("CLEAR10", 0, 3300, 550),
+        ("CLEAR100", 0, 9964, 5000),
     ],
 )
 def test_clear_data_module(tmpdir, dataset_name, chunk_id, num_tr, num_te):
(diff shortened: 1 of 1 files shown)


test/renate/data/test_data_module.py
```diff
@@ -91,15 +91,15 @@ def test_torchvision_data_module(tmpdir, dataset_name, num_tr, num_te, x_shape):
 @pytest.mark.parametrize(
     "dataset_name,chunk_id,num_tr,num_te",
     [
-        ("CLEAR10", 0, 2986, 500),
-        ("CLEAR100", 0, 9945, 4984),
+        ("CLEAR10", 0, 3300, 550),
+        ("CLEAR100", 0, 9964, 5000),
```
wistuba, line 95: did you verify that the first number is not 10K but 9964?

test/renate/data/test_data_module.py
```diff
@@ -131,7 +131,7 @@ def test_tiny_imagenet_data_module(tmpdir):
     assert isinstance(val_data, Dataset)
     assert len(test_data) == num_te
     assert isinstance(test_data, Dataset)
-    assert train_data[0][0].size() == test_data[0][0].size() == (3, 64, 64)
+    assert train_data[0][0].size == test_data[0][0].size == (64, 64)
```
wistuba, line 134: can we drop the second `[0]` and check for `(3, 64, 64)`?

test/renate/data/test_data_module.py
prabhuteja12: Yes.

test/renate/data/test_data_module.py
prabhuteja12: No. `train_data[0]` is the first dataset element, and `train_data[0][0]` is a PIL Image of size 64x64.

wistuba (approved):

PR #447, opened 2023-10-05
Core50 dataset

prabhuteja12:
This PR enables the use of CoRe50 dataset. 

### DIFF
--- a/benchmarks/experiment_configs/datasets/core50.json
+++ b/benchmarks/experiment_configs/datasets/core50.json
@@ -1,0 +1,4 @@
+{
+    "dataset_name": "Core50",
+    "num_outputs": 50
+}
--- a/src/renate/benchmark/datasets/vision_datasets.py
+++ b/src/renate/benchmark/datasets/vision_datasets.py
@@ -569,7 +569,7 @@ class CORE50DataModule(DataIncrementalDataModule):
         self._train_data, self._val_data = self._split_train_val_data(train_data)
         self._test_data = ImageDataset(*self._parse_file_lists("test"))
 
-    def _parse_file_lists(self, stage: str) -> Tuple[str, int]:
+    def _parse_file_lists(self, stage: str) -> Tuple[List[str], List[int]]:
         data_id = self.data_id if stage == "train" else -1
         idx_list = self._LUP[self._scenario][0][data_id]
         X = [os.path.join(self._complete_data_path, self._paths[idx]) for idx in idx_list]
--- a/src/renate/benchmark/datasets/vision_datasets.py
+++ b/src/renate/benchmark/datasets/vision_datasets.py
@@ -3,7 +3,8 @@
 import json
 import os
 from pathlib import Path
-from typing import List, Optional, Tuple, Union
+import pickle
+from typing import List, Literal, Optional, Tuple, Union
 
 import gdown
 import pandas as pd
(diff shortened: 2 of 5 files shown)


benchmarks/experiment_configs/datasets/core50.json
```diff
@@ -0,0 +1,4 @@
+{
+    "dataset_name": "Core50",
```
wistuba, line 2: do we want to refer to it by this name or "CORe50"?

wistuba (changes requested):
can you please add it to the list here?
https://renate.readthedocs.io/en/latest/benchmarking/renate_benchmarks.html#datasets

src/renate/benchmark/datasets/vision_datasets.py
```diff
@@ -471,3 +472,106 @@ def setup(self) -> None:
...
+        with open(os.path.join(self._data_path, self._dataset_name, "labels.pkl"), "rb") as f:
+            self._labels = pickle.load(f)
+
+    def setup(self) -> None:
+        """Make assignments: train/test splits (CORe50 dataset only has train and test splits)."""
+        train_data = ImageDataset(*self._parse_file_lists("train"))
+        self._train_data, self._val_data = self._split_train_val_data(train_data)
+        self._test_data = ImageDataset(*self._parse_file_lists("test"))
+
+    def _parse_file_lists(self, stage: str) -> Tuple[str, str]:
```
wistuba, line 571: wrong return typing

src/renate/benchmark/datasets/vision_datasets.py
```diff
@@ -471,3 +472,106 @@ def setup(self) -> None:
...
+                    file_name,
+                )
+        with open(os.path.join(self._data_path, self._dataset_name, "paths.pkl"), "rb") as f:
+            self._paths = pickle.load(f)
+
+        with open(os.path.join(self._data_path, self._dataset_name, "LUP.pkl"), "rb") as f:
+            self._LUP = pickle.load(f)
+
+        with open(os.path.join(self._data_path, self._dataset_name, "labels.pkl"), "rb") as f:
+            self._labels = pickle.load(f)
```
wistuba, line 563: this should be in `setup`. otherwise, it will not be in memory when we access the attributes in `setup`.

prabhuteja12:
> do we want to refer to it by this name or "CORe50"? 

I chose this only to avoid spelling mistakes.

benchmarks/experiment_configs/datasets/core50.json
prabhuteja12: I chose this only to avoid spelling mistakes.

src/renate/benchmark/datasets/vision_datasets.py
prabhuteja12: Changing to `Tuple[str, int]`

wistuba: I don't think this is correct. the first entry of the tuple is List[str]

wistuba (approved):

wistuba (changes requested):
can you please add it to the list here?
https://renate.readthedocs.io/en/latest/benchmarking/renate_benchmarks.html#datasets

wistuba (approved):

PR #448, opened 2023-10-06
Logging additional metrics

prabhuteja12:

### DIFF
--- a/src/renate/updaters/model_updater.py
+++ b/src/renate/updaters/model_updater.py
@@ -63,10 +63,14 @@ class SyneTuneCallback(Callback):
         )
 
     def on_train_epoch_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
+        self._additional_metrics.on_train_epoch_end()
         self._log(trainer=trainer, pl_module=pl_module)
 
     def on_validation_epoch_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
         self._log(trainer=trainer, pl_module=pl_module)
+
+    def on_train_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
+        self._additional_metrics.on_train_start()
 
 
 class RenateModelCheckpoint(ModelCheckpoint):
--- a/src/renate/updaters/model_updater.py
+++ b/src/renate/updaters/model_updater.py
@@ -57,7 +57,7 @@ class SyneTuneCallback(Callback):
         to_report = {k: v.item() for k, v in trainer.logged_metrics.items()}
         to_report.update(self._additional_metrics(pl_module))
         self._report(
-            **{k: v.item() for k, v in trainer.logged_metrics.items()},
+            **to_report,
             step=trainer.current_epoch,
             epoch=trainer.current_epoch + 1,
         )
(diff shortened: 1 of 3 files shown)


src/renate/updaters/model_updater.py
```diff
@@ -40,28 +41,36 @@ def __init__(self, val_enabled: bool):
...
     def on_train_epoch_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
-        self._log(trainer=trainer, training=pl_module.training)
+        self._additional_metrics.on_train_epoch_end()
+        self._log(trainer=trainer, pl_module=pl_module)
 
     def on_validation_epoch_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
-        self._log(trainer=trainer, training=pl_module.training)
+        self._log(trainer=trainer, pl_module=pl_module)
+
+    def on_train_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
```
wistuba, line 72: is this called also for sanity_checking?

src/renate/utils/misc.py
```diff
@@ -37,3 +39,45 @@ def maybe_populate_mask_and_ignore_logits(
...
+
+    def on_train_start(self) -> None:
+        self._training_times.append(time.time())
+
+    def on_train_epoch_end(self) -> None:
+        self._training_times.append(time.time())
+
+    def __call__(self, model: torch.nn.Module) -> Dict[str, Union[float, int]]:
+        if self._training_times:
+            epoch_training_time = self._training_times[-1] - self._training_times[0]
```
wistuba, line 56: why is this the epoch_training_time? isn't this total training time?

src/renate/utils/misc.py
```diff
@@ -37,3 +39,45 @@ def maybe_populate_mask_and_ignore_logits(
...
+
+class AdditionalTrainingMetrics(Callback):
+    def __init__(self) -> None:
+        self._training_times = []
+
+    def on_train_start(self) -> None:
+        self._training_times.append(time.time())
+
+    def on_train_epoch_end(self) -> None:
+        self._training_times.append(time.time())
```
wistuba, line 52: instead of using a list, maybe simply save start and end time?

src/renate/updaters/model_updater.py
prabhuteja12: What is `this`?

src/renate/utils/misc.py
prabhuteja12: I just wanted to keep the flexibility of retrieving all times, though the current interface doesn't
necessarily allow for it.

src/renate/utils/misc.py
prabhuteja12: Modifying.

src/renate/updaters/model_updater.py
wistuba: `on_train_start()`

src/renate/utils/misc.py
wistuba: unless we use it or plan to use it soon, we shouldn't sacrifice simplicity for flexibility. unless you strongly disagree, let's keep this simple.

src/renate/updaters/model_updater.py
prabhuteja12: It runs on on_validation_start, hence the special handling checking for None in the Additional metrics class.

